PR #11, opened 2015-06-11
Adding Ros Jade

ruffsl:
I'd like to pull in new Jade images into the docker repo.
@tfoote , check that this all look good.

I've kept ubuntu to trusty, as I think keeping the ROS images on LTS of ubuntu (and keeping the "latest" tag on LTS of ROS, indigo) would be the best move.


ruffsl:
By the way, what is the exact EOL date for indigo, 2017-04-26, same as Precise Pangolin?
I thought indigo was on trusty's cycle. I'd like to note it for both indigo and jade in the updated docker-library manafest.
http://wiki.ros.org/Distributions/Timeline
http://en.wikipedia.org/wiki/List_of_Ubuntu_releases

tfoote:
+1 from quick review

ruffsl:
Correction,
EOL for 
indigo/trusty LTS: 2019-04
jade: ~2 yrs after initial release, so 2017-05

PR #23, opened 2015-10-07
Added catkin-make dependancy: build-essentials

rethink-imcmahon:
`catkin_make` fails on a basic catkin workspace unless `build-essentials` is installed.

### DIFF
--- a/ros/indigo/indigo-ros-core/Dockerfile
+++ b/ros/indigo/indigo-ros-core/Dockerfile
@@ -20,6 +20,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     python-rosdep \
     python-rosinstall \
     python-vcstools \
+    build-essential \
     && rm -rf /var/lib/apt/lists/*
 
 # bootstrap rosdep


tfoote:
I'll have to think about this. If it's really needed we should consider adding it as a dependency on the package instead of just requiring it be there.

scpeters:
It depends on the purpose of the image. If it is just meant for running ROS code, then `build-essentials` isn't required. Do we want to a have very compact image intended for running code and then supply an image with development tools as well?

rethink-imcmahon:
Those are good points, guys. This PR was my knee-jerk reaction to not being able to run `catkin_make` despite `catkin` being installed. Perhaps we need a second set of Docker images, one for Dev and one for Production? I imagine Production wouldn't need `catkin` at all.

ruffsl:
I agree with @tfoote , that if this warrants enough need, then it may be good to induce as a dependency on the package. Perhaps it could be induced as an optimal or recommended dependency? If I remember correctly, I had thought the Ubuntu image has apt-get set to install recommended by default [[1]](http://askubuntu.com/a/117646) [[2]](https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html#s-priority).

varhub:
Getting up this old topic, how many size is required for devel inside?
In any case, every apt-get install without a apt-get clean+autoclean adds .deb forever. So it should be reviewed.

ruffsl:
@varhub , sorry, not sure of what you mean? Are you speaking of the [official ros](https://hub.docker.com/_/ros/) images and the size of a resulting devel enabeled image?

ruffsl:
Closing for in part from inactivity. 

If you'd like a ros image for compiling with build tools pre-installed, I'd suggest making an automated build repo on dockerhub, and having the dockerfile build `FROM` one of the official ROS tags on docker library. This also have it trigger to rebuild whenever the official ROS repo churns from new Ubuntu base images. That way you can compile from an up-to-date docker image, while customizing the build/dev dependencies to your liking.

varhub:
Yes @ruffsl  I was speaking about official ros images. This `/var/lib/apt/lists/*` is useless in terms of space wasting.
I created https://github.com/osrf/docker_images/pull/74 explaining rationale.

PS: sorry for late response

PR #24, opened 2015-10-08
Touching up Makefiles for ROS images

ruffsl:
Adding osrf/ to repo name in Makefile to match Dockerfiles and removing tag and push, as we have automated build repos now.


tfoote:
Thanks. lgtm

PR #25, opened 2015-11-16
Fixing entry point script

ruffsl:
removing quotes to avoid exec attempting to find a command named as the entire string, braking commands with given arguments.


ruffsl:
Updating ros docs on dockerhub, ran into this issue when using commands with docker-compose.

PR #45, opened 2017-01-27
Adding package_version checking for Gazebo

ruffsl:
if package_version is left empty in the platform.yaml
the create_docker.py script with check OSRF's package index for it instead
Saves maintainers like me from missing version bumps

Let me know if you have any suggestions to do this better.


ruffsl:
@j-rivero or @tfoote , any suggestions or does this look good? I'd prefer to use HTTPS, or some authenticated. I'm thinking of trying out this [Python APT Library](http://apt.alioth.debian.org/python-apt-doc/library/index.html), to use the pagage manager for the authenticated idex checking and package version lookup.

gazebo/create_dockerfiles.py
```diff
@@ -16,6 +19,35 @@
 from ros_buildfarm.docker_common import OrderedLoad
 
 
+def checkVersion(data):
+    """Checks current package_version from online package index"""
+
+    # Determine URL
+    urlTemplate = string.Template(
+        """
+        http://packages.osrfoundation.org/gazebo/$os_name/dists/$os_code_name/main/binary-$arch/Packages
```
j-rivero: It just works but I would prefer to use the `$os_name-stable` to name the stable repositories. Minor thing.

gazebo/create_dockerfiles.py
```diff
@@ -16,6 +19,35 @@
...
+
+    # Determine seach pattern
+    patternTemplate = string.Template(r'(\bPackage: gazebo$version\n)(.*\n)')
+    pattern_raw = patternTemplate.substitute(data)
+    pattern = re.compile(pattern_raw)
+
+    # Parse for version_number
+    matchs = re.search(pattern, package_index)
+    version_line = matchs.groups(0)[1] # Grab the second line of the first match
+    version_number = re.search(r'\d{1}[.]\d{1}[.]\d{1}', version_line).group(0) # extract version_number
```
j-rivero: It's true that we keep a weird version name (`$version + ~$distro`) but we probably need to cover more use cases than just the standard `d.d.d`. I would suggest to get all the field and exclude just (and only) the final ~ and what comes next until the end. I.e: `8.0.0~beta4~yakkety` to be `8.0.0~beta4`.

You can try it replacing the stable URL above by `$os_name-prerelease`

gazebo/create_dockerfiles.py
ruffsl: What vocab would you use to call `stable/prerelease`? 
I'll make an alias for it so it could be `$os_name-$sothing`

gazebo/create_dockerfiles.py
ruffsl: I'm a little rusty on my regex, whats a good way to grab that, up to the second tilda?
Would this cover all your cases?
`\d{1}[.]\d{1}[.]\d{1}~\w*~`
or
`((\d)+[.])+\d+((~|\+)[\w,\-]*(?=~|\n)|\n)`

The last one should match with stuff like:
`127.0.0.1~beta9000~spam\n` -> `127.0.0.1~beta9000`
`127.0.0.1~beta9000\n` -> `127.0.0.1~beta9000`
`127.0.0\n` -> `127.0.0`
I'm not sure of all numeric conventions there are in package versioning format constraints?
Theres a wide variaty of stuff like stuff like [this](http://packages.osrfoundation.org/gazebo/ubuntu-stable/dists/trusty/main/binary-amd64/Packages):
`Version: 0.0.0+git20170131r43c21ae5115b693ec0edc551abbc189a54e3c2ab-4~trusty` -> `0.0.0+git20170131r43c21ae5115b693ec0edc551abbc189a54e3c2ab-4`

FYI: you can debug them with [regexr.com](http://regexr.com/) or [debuggex.com](https://www.debuggex.com/)

ruffsl: Looking for the leading `Version: ` sting and tailing `~distro\n` might be easier,
but I'm not sure if they all have the tailing `~distro\n` string or not.

`(?!Version\:\s)([\d|\w|\+|.|\-]+)(?=(~\w+\n))`

gazebo/create_dockerfiles.py
j-rivero: `repo_type`? `builds_type`? Not easy.

gazebo/create_dockerfiles.py
j-rivero: We have released all the packages with the `~distro\n` until today. I don't have plans to change it.

PR #51, opened 2017-05-11
add lunar Dockerfiles

mikaelarguedas:
This adds the dockerfiles to create official ROS lunar images on dockerhub.

Note that the images_yaml.em Empy template is that same for all distro and thus could be shared but I prefer to do that in a follow-up PR

### DIFF
--- a/ros/indigo/images.yaml.em
+++ b/ros/indigo/images.yaml.em
@@ -36,7 +36,7 @@ images:
         ros_packages:
             - perception
     @(rosdistro_name)-desktop:
-        base_image: @(user_name):@(rosdistro_name)-ros-base
+        base_image: @(user_name):@(rosdistro_name)-robot
         maintainer_name: @(maintainer_name)
         template_name: docker_images/create_ros_image.Dockerfile.em
         template_packages:
--- a/ros/indigo/indigo-desktop-full/Dockerfile
+++ b/ros/indigo/indigo-desktop-full/Dockerfile
@@ -5,6 +5,6 @@ FROM osrf/ros:indigo-desktop
 
 # install ros packages
 RUN apt-get update && apt-get install -y \
-    ros-indigo-desktop-full=1.1.4-0* \
+    ros-indigo-desktop-full=1.1.5-0* \
     && rm -rf /var/lib/apt/lists/*
 
--- a/ros/indigo/indigo-desktop-full/Dockerfile
+++ b/ros/indigo/indigo-desktop-full/Dockerfile
@@ -1,6 +1,6 @@
 # This is an auto generated Dockerfile for ros:indigo-desktop-full
 # generated from templates/docker_images/create_ros_image.Dockerfile.em
-# generated on 2017-01-27 02:33:52 +0000
+# generated on 2017-05-23 06:51:52 +0000
 FROM osrf/ros:indigo-desktop
 
 # install ros packages
--- a/ros/indigo/indigo-desktop/Dockerfile
+++ b/ros/indigo/indigo-desktop/Dockerfile
@@ -5,6 +5,6 @@ FROM ros:indigo-ros-base
 
 # install ros packages
 RUN apt-get update && apt-get install -y \
-    ros-indigo-desktop=1.1.4-0* \
+    ros-indigo-desktop=1.1.5-0* \
     && rm -rf /var/lib/apt/lists/*
 
--- a/ros/indigo/indigo-desktop/Dockerfile
+++ b/ros/indigo/indigo-desktop/Dockerfile
@@ -1,7 +1,7 @@
 # This is an auto generated Dockerfile for ros:indigo-desktop
 # generated from templates/docker_images/create_ros_image.Dockerfile.em
-# generated on 2017-01-27 02:33:52 +0000
-FROM ros:indigo-ros-base
+# generated on 2017-05-23 06:51:52 +0000
+FROM ros:indigo-robot
 
 # install ros packages
 RUN apt-get update && apt-get install -y \
--- a/ros/indigo/indigo-perception/Dockerfile
+++ b/ros/indigo/indigo-perception/Dockerfile
@@ -5,6 +5,6 @@ FROM ros:indigo-ros-base
 
 # install ros packages
 RUN apt-get update && apt-get install -y \
-    ros-indigo-perception=1.1.4-0* \
+    ros-indigo-perception=1.1.5-0* \
     && rm -rf /var/lib/apt/lists/*
 
--- a/ros/indigo/indigo-perception/Dockerfile
+++ b/ros/indigo/indigo-perception/Dockerfile
@@ -1,6 +1,6 @@
 # This is an auto generated Dockerfile for ros:indigo-perception
 # generated from templates/docker_images/create_ros_image.Dockerfile.em
-# generated on 2017-01-27 02:33:52 +0000
+# generated on 2017-05-23 06:51:52 +0000
 FROM ros:indigo-ros-base
 
 # install ros packages
--- a/ros/indigo/indigo-robot/Dockerfile
+++ b/ros/indigo/indigo-robot/Dockerfile
@@ -5,6 +5,6 @@ FROM ros:indigo-ros-base
 
 # install ros packages
 RUN apt-get update && apt-get install -y \
-    ros-indigo-robot=1.1.4-0* \
+    ros-indigo-robot=1.1.5-0* \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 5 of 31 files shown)


ruffsl:
All image tags are building successfully with the make file, so LGTM.
Are we planning on sending a PR to [docker-library/official-images](https://github.com/docker-library/official-images) and asking them wait and merge/build/release May 23rd with a last release sync, or just wait to make the PR until then?

mikaelarguedas:
I was thinking to wait until then. I don't know how responsive they usually are, do you think it's better to open it beforehand?

ruffsl:
They have been quick to review our PRs for ROS & Gazebo, as our templates are static/consistent. The turnaround is about a day or two, waiting for the admins to trigger a test build then check it. I'd say we could save it until the release sync is ready, to ensure its all included in the dockerhub images.

mikaelarguedas:
Sounds good, I'll do it a few days before the official release date then, thanks for the info.
Not really related but: is it on purpose that we provide `robot` and `perception` images but not `desktop` and `desktop-full` ?

tfoote:
For official docker images the graphical dependencies are too much of a security risk to be accepted. But we can host them in the osrf organization.

mikaelarguedas:
It makes sense, thanks for the explanation. I think it would be valuable indeed to have a reduced installation time when using the docker images than starting from a fresh ubuntu image. But is it worth hosting the images in two different places? I'm not sure

ruffsl:
The official dockerhub library hosts the `core`, `base`, `robot`, and `perception` tags, while OSRF just hosts the `desktop` tag that builds from the official's `base` tag, and the `desktop-full` tag that builds from the former. If a user pulls `osrf/ros:lunar-desktop` after pulling `ros:lunar`, then the docker host can bypass fetching the first most common layers shared among `base` and `desktop` tags, so it's not like the two repos are entirely different and would use unnecessary disk space.

Also, once OSRF simply adds the new distro tags to OSRF's automated dockerhub repo, little else is ever required with dockerhub (they will be triggered to rebuild automatically when upstream images are triggered), other than keeping the package versions in the dockerfiles up-to-date this github repo, which is required for maintaining the official image anyhow via scripts, so the additional tags incur little overhead.

~~My only gripe is that `desktop-full` can be one build phase behind, as using the  `Repository Links` feature will trigger all tags to build simultaneously, thus the new `desktop-full` tag will pull/build from the old `desktop` tag before the new `desktop` is built/pushed. Nvidia previously resorted to using [Gitlabs CI pipeline](https://github.com/NVIDIA/nvidia-docker/issues/18#issuecomment-273346844) to achieve the similar nested builds that official docker repo tags afford natively.~~
Nevermind, haven't checked this in a while, but perhaps DockerHub has resolved this recently, wither the effect is intermittent or it seems to be working now. Wish I could find the ticketed issue stating if this build behavior was fixed.

```
$ docker pull ros
Using default tag: latest
latest: Pulling from library/ros
...
fe56bbca99ef: Pull complete 
Digest: sha256:baf314578a1ced9a8f55a1c496648cb692ce8203401a0a41c9eb15bec17db2ce
Status: Downloaded newer image for ros:latest

$ docker pull osrf/ros:kinetic-desktop
kinetic-desktop: Pulling from osrf/ros
...
fe56bbca99ef: Already exists 
d69693254fdd: Pull complete 
Digest: sha256:5ce08cec976c46bd6467e3caae84983c99acf3d85e57b8c154fcfad5b3518099
Status: Downloaded newer image for osrf/ros:kinetic-desktop

$ docker pull osrf/ros:kinetic-desktop-full
kinetic-desktop-full: Pulling from osrf/ros
...
fe56bbca99ef: Already exists 
d69693254fdd: Already exists 
6f0eb47bdc79: Pull complete 
Digest: sha256:06979d445c807dc28d5da32871d55ee01fc9a3e3675c1509fed1d757d084c571
Status: Downloaded newer image for osrf/ros:kinetic-desktop-full
```

mikaelarguedas:
@ruffsl I fixed the missing locales in https://github.com/osrf/docker_templates/pull/13 and regenerated the lunar dockerfiles accordingly. Im going to merge this and make a PR to [docker-library/official-images](https://github.com/docker-library/official-images). I think that the other dockerfiles should also be regenerated to fixe the locales and target the latest version of ros metapackages. Should I do all in a single PR to upstream `official-images` or is it better practice to separate "updating Dockerfiles" from "adding new images" ?

mikaelarguedas:
@ruffsl Just noticed https://github.com/osrf/docker_images/pull/52 and https://github.com/osrf/docker_templates/pull/12. I did add the install step of locales in https://github.com/osrf/docker_templates/pull/13. I saw you take a different approach by not adding it to the template but to the docker images dependencies. I thought that any template calling locale-gen should install local explicitely rather than relying on the dependency being listed externally. Please let me know if this doesn't make sense and I'll revert it to mimic the approach you took in #52

ruffsl:
> I thought that any template calling locale-gen should install local explicitely rather than relying on the dependency being listed externally. Please let me know if this doesn't make sense and I'll revert it to mimic the approach you took in #52

@mikaelarguedas, indeed I thought it would be best to add the dependency into the yaml config then set the variables after apt install, rather than hardcoding them into the templates. Although lots of ros/gazebo code most likely requires locals, we can avoid an expensive/unnecessary apt-update, shaving a minute or so off each build time, by grouping in that dependency into the yaml configs.

Also, remember that these templates are not necessary Ubuntu specific. Other users can swap the base image at will, so assuming a package names, version numbers (or lack of) may impede its use on other base images.

My hope would be that the core metapackages upstream would declare locals as dependency, so that we wouldn't have to, ( thoughts, @tfoote , @j-rivero ?)  For now, the least intrusive move I think would be to mimic what I did for ros2. By the way, should we add `LC_ALL`, like we have `LANG`, to all templates?

mikaelarguedas:
> Also, remember that these templates are not necessary Ubuntu specific. Other users can swap the base image at will, so assuming a package names, version numbers (or lack of) may impede its use on other base images.

Agreed, as a side note: both Debian and Ubuntu don't ship with locales and they are our supported platforms. We don't test or build packages for any other distributions.

> indeed I thought it would be best to add the dependency into the yaml config then set the variables after apt install

Do we agree that means that any Dockerfile based on this template will fail to build if the configuration file doesn't have `locales` listed explicitly as a dependency ?
I'm fine with both solutions I'm just trying to evaluate which approach is most likely to bite us in the future.

> My hope would be that the core metapackages upstream would declare locales as dependency, so that we wouldn't have to

@dirk-thomas any thoughts on this ?

dirk-thomas:
> > My hope would be that the core metapackages upstream would declare locales as dependency, so that we wouldn't have to

> dirk-thomas any thoughts on this ?

The ROS packages can't set the current `locale` - only have a dependency on the `locales` package. But the dependency is worthless imo without the right selected `locale`. Therefore I think this needs to be handles within the Dockerfiles.

ruffsl:
> Do we agree that means that any Dockerfile based on this template will fail to build if the configuration file doesn't have locales listed explicitly as a dependency ?

Hmm, good point. Similarly, the Dockerfile would fail to build if the config was such that rosdep was not installed. Oh, better thought! A exelent compromise I think would be to add locals to that same apt-get section:
https://github.com/osrf/docker_templates/blob/e96e12867acf4096195fca36ccac236ab093ce1a/ros_docker_images/templates/docker_images/create_ros_core_image.Dockerfile.em#L40-L45


Perhaps a similar section could be added to Gazebo's template, although it's not currently setting `locale-gen`, `LANG`, `LC_ALL` etc.

mikaelarguedas:
Sounds good, I'll move back the locales related commands to after the installation of these packages then. Do you remember why moving them was necessary in the first place (not sure what "suggestions" you are referring to in [this comment](https://github.com/osrf/docker_templates/pull/12#issuecomment-301886133))?

ruffsl:
> Do you remember why moving them was necessary

The `locale-gen` command, i.e using docker RUN as opposed to being an ENV, is not on system, and so we moved it to after the locals package was installed. I also moved those ENV lines with it, as I wanted to keep them all together with the relevant command.

ruffsl:
> Should I do all in a single PR to upstream official-images or is it better practice to separate "updating Dockerfiles" from "adding new images" ?

Just one PR with the updated and new ROS Dockerfiles should be fine, think we did this with kinetic. We should make a separate PR for Gazebo though, if we end up updating the templates for those images as well.

mikaelarguedas:
Ok I updated the Dockerfiles to use the installed locales from https://github.com/osrf/docker_templates/pull/14
I bumped the packages version number for idigo (note thhat I havent done it for jade and kinetic because the sync is still pending) 
I also modified the desktop images to pull from robot because every ROS distro as of Indigo defines the desktop variant as an extension of the robot metapackage (see http://www.ros.org/reps/rep-0142.html)

mikaelarguedas:
I tested all of them locally and they built successfully, so I'm going to merge this to get the pull request to get these images upstream going

ruffsl:
> I also modified the desktop images to pull from robot because every ROS distro as of Indigo defines the desktop variant as an extension of the robot metapackage (see http://www.ros.org/reps/rep-0142.html)

Nice find, that originally escaped me. Much better reuse of image layers. I'll have to make a PR to update the official doc page.

PR #56, opened 2017-06-07
Updating gazebo 7 and 8

nkoenig:

PR #58, opened 2017-06-10
bump kinetic version to 1.3.1

mikaelarguedas:
now the the kinetic sync happened we can update the docker image. Successfully built all of them locally

### DIFF
--- a/ros/kinetic/kinetic-desktop-full/Dockerfile
+++ b/ros/kinetic/kinetic-desktop-full/Dockerfile
@@ -1,10 +1,10 @@
 # This is an auto generated Dockerfile for ros:kinetic-desktop-full
 # generated from templates/docker_images/create_ros_image.Dockerfile.em
-# generated on 2017-05-23 21:41:39 +0000
+# generated on 2017-06-10 20:30:32 +0000
 FROM osrf/ros:kinetic-desktop
 
 # install ros packages
 RUN apt-get update && apt-get install -y \
-    ros-kinetic-desktop-full=1.3.0-0* \
+    ros-kinetic-desktop-full=1.3.1-0* \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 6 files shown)


ruffsl:
LGTM

By the way, do you think it would be better to swap the create_dockerfiles script for ros to scrape from the repo package index like gazebo does rather than the get_index function from rosdistro python lib. I never knew the apt repo and the rosdistro index would be out of sync for so long.

mikaelarguedas:
We try to have syncs every two weeks but if there are regression it can take longer. I think it's a good feature to have (scraping from the repos rather than rosdistro) :+1: .

Hopefully pushing wrong version numbers should not happen again. We discussed it within the ROS team and the conclusion (https://github.com/ros/metapackages/issues/16#issuecomment-303518969) is that updating the images of a ROS distro should be the responsability of the ROS boss of that distro. So bumping the version number of a distro you are not responsible of should not happen in theory.
Furthermore these images should not have to be updated very often given that they need to change only when the metapackages change. Which is pretty/very rare

ruffsl:
Ok, I'll work on a PR to enable the ros script to scrape from the repos.
Could you make the PR change to official library, thanks!

mikaelarguedas:
> Ok, I'll work on a PR to enable the ros script to scrape from the repos.

Awesome thanks! ideally we could have an option to select the test (ros-shadow-fixed) or the main repo so that the "testing" image would stay in sync with rosdistro while the "stable" one would be updated every sync

Sure, I'll wait for @nkoenig response to https://github.com/osrf/docker_images/pull/56#issuecomment-307600477 to include gazebo7.8 in it or not. Thanks for reviewing!

mikaelarguedas:
Given that the changes are not related and that one changes the way to find os version while this one is just bumping a version number I'm going to open a self contained PR only for this change and we will open another one for the gazebo changes and iterate on it if the new strategy to detect version is not good practice.

PR #59, opened 2017-06-16
Changing ROS package version lookup

ruffsl:
Now scrapping version info for ros package using online package index instead of rosdistro lookups.
Version info lookup can also be bypassed by setting `version` field in platform.yaml to something other than empty.


ruffsl:
Perhaps eventually we could think about version introspection of releases for beyond debian as well.

mikaelarguedas:
@ruffsl can this branch be deleted?

PR #62, opened 2017-07-06
Updating keyserver URL

alek5k:
When I tried to duplicate the dockerfile commands in my container I wasn't able to add the key from the keyserver. See error from my container below:

```
root@f19b33dcdafa:/# apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys 421C365BD9FF1F717815A3895523BAEEB01FA116
Executing: /tmp/tmp.RCcSlVBGYW/gpg.1.sh --keyserver
ha.pool.sks-keyservers.net
--recv-keys
421C365BD9FF1F717815A3895523BAEEB01FA116
gpg: requesting key B01FA116 from hkp server ha.pool.sks-keyservers.net

gpg: keyserver timed out
gpg: keyserver receive failed: keyserver error
```
Based on the ROS installation tutorial found at this URL:
http://wiki.ros.org/kinetic/Installation/Ubuntu

The recommendation is to use the url:
hkp://ha.pool.sks-keyservers.net:80

This pull request changes the keyserver url in the dockerfiles to use the tutorial URLs.

After making this change, adding the keyserver to my container was working.

### DIFF
--- a/gazebo/gazebo4/gzserver4/Dockerfile
+++ b/gazebo/gazebo4/gzserver4/Dockerfile
@@ -4,7 +4,7 @@
 FROM ubuntu:trusty
 
 # setup keys
-RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys D2486D2DD83DB69272AFE98867170598AF249743
+RUN apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-keys D2486D2DD83DB69272AFE98867170598AF249743
 
 # setup sources.list
 RUN echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-latest.list
--- a/gazebo/gazebo5/gzserver5/Dockerfile
+++ b/gazebo/gazebo5/gzserver5/Dockerfile
@@ -4,7 +4,7 @@
 FROM ubuntu:trusty
 
 # setup keys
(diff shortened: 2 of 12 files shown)


alek5k:
Just noticed the dockerfiles are autogenerated, where can this be updated?

ruffsl:
Interesting, specifying the use of HTTP Keyserver Protocol via port 80 resolves your issue? Is this something to do with your network environment, firewall, or IP setup. I've noticed some issues before with fetching keys from the keyserver when my system was forced IPv4 (or IPv6?). 

Is there another best practice reference that recommends hkp via port 80 instead of the default port (I think 11371)? We can change this in the [template repo](https://github.com/osrf/docker_templates/blob/master/ros_docker_images/templates/docker_images/create_ros_core_image.Dockerfile.em#L28) if need be. @tfoote , any opinion?

tfoote:
Apparently sometimes it works better for people with firewalls: https://github.com/ros-infrastructure/roswiki/issues/140

alek5k:
If you see [this Ubuntu package](https://launchpad.net/ubuntu/+source/software-properties/0.78.1) update, the same changes were implemented.

I have made a new [pull request](https://github.com/osrf/docker_templates/pull/18) to the templates repo.

ruffsl:
I ran this issue by the docker library maintainers, their responce:
https://botbot.me/freenode/docker-library/2017-07-07/?msg=88282135&page=1

I think using the p80.pool via port 80, i.e `hkp://p80.pool.sks-keyservers.net:80` would resolve the firewall issues. https://sks-keyservers.net/overview-of-pools.php

> **p80.pool.sks-keyservers.net**
This is a pool containing only servers available on port 80 (needs to be used as hkp://p80.pool.sks-keyservers.net:80)

PR #63, opened 2017-07-07
Keyserver fix

ruffsl:
Using new template with improved keyserver url
and removed timestamp for auto dockerfile build diff detection.


mikaelarguedas:
@ruffsl can this branch be deleted?

PR #65, opened 2017-08-12
Adding Multi OS support for ROS Dockerfiles

ruffsl:
Adding debian dockerfiles for ROS1
With additional dockerfolder auto generation from central manifests
and travis support


mikaelarguedas:
@ruffsl A few questions: How was the set of distribution targeted selected ? stable releases for Ubuntu and Debian ?
Would there be a way to update this set automatically based on what the buildfarm is building? (thinking about jade going eol soon and missing platforms for lunar).
Looks like CI failed multiple times because of apt errors. Should we use the apt wrapper provided by ros_buildfarm in these files to avoid false positives>

ruffsl:
@tfoote and I talked a little offline and wanted to check out docker hub's milti-arch features for our ROS repo with supported architectures. While we were at it, @tfoote suggested we also add support for our other supported OS, including non LTS. I'm still a bit hesitant to try and support every supported OS/arch/distro, so I settled with just focusing on automating for LTS OS' and see how soothly go from there.

I've automated generating the manifest file from this repo for ROS that would be used in the PR for the official library upstream, as well as the folder generation for the sole manifest and deriving templates. PR forthcoming.

Yea, debian's server seem being rather flaky today, but I'm not sure it would justify wrapping all our apt calls in the dockerfile. I think it may not be as transparent, and would add more dependencies of the base images.

mikaelarguedas:
Cool!

> While we were at it, @tfoote suggested we also add support for our other supported OS, including non LTS. I'm still a bit hesitant to try and support every supported OS/arch/distro, so I settled with just focusing on automating for LTS OS' and see how soothly go from there.

Yeah if it works well we can always add the non-LTS later. That would save me a ton of time these days

> Yea, debian's server seem being rather flaky today, but I'm not sure it would justify wrapping all our apt calls in the dockerfile. I think it may not be as transparent, and would add more dependencies of the base images.

Yeah it''s been very unstable for [a while](https://github.com/ros-infrastructure/ros_buildfarm/issues/455). [Cloudfront](https://cloudfront.debian.net/) seemed to be more stable / up-to-date / responsive to issues if it's of any help. It may still be worth wrapping it because I would expect a lot of false positive on the bigger images (desktop or desktop-full).

PR #66, opened 2017-08-14
Adding support to ROS repo for Multiple Architectures on Dockerhub

ruffsl:


mikaelarguedas:
Do we expect the ROS image.yaml templates to be different in some scenarios ? Or could we consider using the same for all ros distributions (ROS variants not really changing over the years)

ruffsl:
@mikaelarguedas , I've enabled the manifest to provide the file to be used as the image.yaml.em, so we could overide the default image templates to different releases, such as if we want to provided a new meta package, or wanted to chage something about the nested tags, etc:
https://github.com/osrf/docker_images/blob/084eba3a61100bb8f89cd0b89f970de09fa4bb6a/ros/manifest.yaml#L9

PR #68, opened 2017-08-22
Updating templates for ROS2

ruffsl:
Goes with https://github.com/osrf/docker_templates/pull/23


mikaelarguedas:
@ruffsl did anybody review this?
I'm surprised by the choice of packages and regex, is there a rationale somewhere as of what goes in what image ?

ruffsl:
Dirk has asked me begin dockerizing ros2 bridge image to use for an internal project, as the target platform does not yet support the same distro as currently with ros2. So I've tagged the images with r2b2 to denote they are only pertinent to the current beta. The posix is just quick way to simply grab every message package, I'm not sure what we need just yet, so we can widdle them down if need be, but they don't add much more space relatively. I also needed to test that build would not time out on docker hub. In any case, ros2 doesn't even have any meta packages yet, so certainly nothing is set in stone and will most likely change as we continue iterating.

As an aside, would it be possible to have a meta package for messages where the only dependcy a dependent packes may have are others in the same meta message package set? I feel like some ros1 msgs packages are declaring in more than may need.

dirk-thomas:
> @dirk-thomas to confirm what we expect in this image and what messages we expect to be able to transmit.

The goal is to communicate joystick commands over the bridge as well as all messages to visualize a robot in rviz.

ros2/r2b2/ubuntu/xenial/images.yaml.em
```diff
@@ -0,0 +1,46 @@
...
+        template_name: docker_images/create_ros2_core_image.Dockerfile.em
+        entrypoint_name: docker_images/ros2_entrypoint.sh
+        template_packages:
+            - docker_templates
+        packages:
+            - python3-pip
+        pip3_install:
+            - argcomplete
+            - flake8
+            - flake8-import-order
```
ruffsl: Should we remove them from the example sros2 dockerfiles then?

ros2/r2b2/ubuntu/xenial/images.yaml.em
```diff
@@ -0,0 +1,46 @@
...
+            - catkin
+            - common-msgs
+            - rosbash
+            - roscpp
+            - roscpp-tutorials
+            - roslaunch
+            - rosmsg
+            - rospy-tutorials
+            - tf2-msgs
+            - .*-*msg
```
ruffsl: Do you know of a way to recursively express patterns? 
`ros-kinetic-[[^-]+-]*msg?` isn't quite valid for catching `ros-kinetic-foo-bar-spam-msgs`

ros2/r2b2/ubuntu/xenial/images.yaml.em
```diff
@@ -0,0 +1,46 @@
...
+        ros2_packages:
+            - ros1-bridge
```
ruffsl: I would say clarity over brevity, as it was not entirely obvious to me which ros1 distro pairs with this ros2 beta, but I'll remove them to avoid any redundancy.

ros2/r2b2/ubuntu/xenial/images.yaml.em
```diff
@@ -0,0 +1,46 @@
...
+            - turtlebot2-*
+            - .*-*msg
```
ruffsl: Same question at above.

ros2/r2b2/ubuntu/xenial/images.yaml.em
mikaelarguedas: not sure how regex are handled in apt but both `ros-kinetic-*msgs` and `ros-kinetic-*-msgs` do match all messages including the ones with several tokens in between e.g. `ros-kinetic-cob-object-detection-msgs` so I would naively expect that the same would work here.

But as per previous review, all the messages we are interested in are already provided by the ros1-bridge in their ros1 and ros2 form so I don't think we need this logic here

ros2/r2b2/ubuntu/xenial/images.yaml.em
mikaelarguedas: In the end I think that a Dockerfile installing only `ros-r2b2-ros1-bridge` will match the requirements here given that it already provides all the packages needed to "communicate joystick commands over the bridge as well as all messages to visualize a robot in rviz."
For convenience we could throw in `ros2run`.

> as it was not entirely obvious to me which ros1 distro pairs with this ros2 beta

Please consider contributing to the [wiki installation page](https://github.com/ros2/ros2/wiki/Linux-Install-Debians) to make it more obvious to users, improvements are always appreciated!

ros2/r2b2/ubuntu/xenial/images.yaml.em
mikaelarguedas: yes, `flake8-import-order` must have been forgotten when all the "development dependencies" have been removed

PR #70, opened 2017-09-11
Version bump for Jade

ruffsl:
Looks like jude got its last sync. I say we let it propagate to the official docker hub tag and then think about how we should eventually remove the tag. I'm thinking for as long as its base image (ubuntu:1604) is getting update.


mikaelarguedas:
Is it on purpose that every commit hash is updated and not only the `jade` ones ?
I guess it's an unrelated change to jade becoming EOL
Otherwise looks good to me

ruffsl:
Yea, the hash changes are only resulted from the commits when restructuring the paths of the files. The other Dockerfiles themselves have not really changed.

mikaelarguedas:
Please delete you branches once PR is merged

PR #71, opened 2017-09-14
Adding EOL examples :turtle:

ruffsl:
For the 10th anniversary of ROS, @tfoote and I created example images for EOL releases of ROS. :turtle: 
This should be helpful for anyone who wants to build or run old archived code, or for general posterity. The targeted releases of Ubuntu are also EOL, so no CI or images hosting is expected here. However, at the time of this PR, they do successfully build, as DockerHub archive those tags of ubuntu, as does the ROS repo for debians.


ruffsl:
These old versions of ubuntu and ros don't seem agree with the current LANG environment config, and give off quite a few warnings. Anyone know what was standard back then?

tfoote:
The lang is hard to say. That's usually only an issue inside the stripped down Docker images, regular installs usually have it all setup at install time.

ruffsl:
Also, are the debs for *boxturtle* still in existence? Just for completeness it would be nice to have, but I did not see them in [packages.ros.org for lucid](http://packages.ros.org/ros/ubuntu/dists/lucid/main/binary-amd64/Packages). Are they truly lost to time?

mikaelarguedas:
I don't think that boxturtle was ever a "real release" more a "beta". AFAICT packages.ros.org starts with cturtle:
http://packages.ros.org/ros/ubuntu/pool/main/r/

ruffsl:
From the wiki archives, it looks like prebuilt binaries existed:  
http://wiki.ros.org/boxturtle/Installation/Ubuntu/Deb

tfoote:
I found backups of the code.ros.org debian packages and have imported boxturtle packages for lucid only. We had packages for hardy, jaunty, karmic, and lucid.

ruffsl:
Added boxturtle; it doesn't have a `setup.bash`, only a `setup.sh`, so I had to tweak an eol entrypoint.

PR #74, opened 2017-09-19
Add full cleanup of apt-get install process (purge cached .deb)

varhub:
* `rm -rf /var/lib/apt/lists/*` only removes package index to ensure a fail
if someone just do an apt-get install without an update (out of date packages)

* but a lot of space is wasted due cached .deb packages:
apt-get install will download .deb files at /var/cache/apt/archives prior to
install them. These "cached" .deb files will be preserved for a while, wasting
docker image size forever and twice (one at this layer, and another one marking
them as deleted on an upper layer when cleanup task is called through cron)

* `apt-get clean` is mandatory to avoid space wasting (docker best practises)

Note (going further): you should avoid `apt-get clean` if package reinstall is
a common step on your workflow. Examples:
  - reinstall of initial version (default settings)
  - downgrade of a package (revert changes)
  - failsafe just if package get wrong (extremely rare)

And you want to run it offline or ensure same version even if package was
dropped from repository.

### DIFF
--- a/gazebo/gazebo4/gzclient4/Dockerfile
+++ b/gazebo/gazebo4/gzclient4/Dockerfile
@@ -8,9 +8,9 @@ RUN apt-get update && apt-get install -q -y \
     mesa-utils \
     module-init-tools \
     x-window-system\
-    && rm -rf /var/lib/apt/lists/*
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # install gazebo packages
 RUN apt-get update && apt-get install -q -y \
     gazebo4=4.1.3-1* \
-    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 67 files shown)


ruffsl:
Thanks @tfoote , I think that this is not immediately obvious, as it's buried deep within the sausage factory in make the offical ubuntu and debian images.
https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105

So I don't think we need this for the modern ubuntu images. I'll patch something similar for our old eol image template though.

varhub:
Yes @tfoote , you are right. This behavior is new for me (I was out of docker for a while) and I got used to do it.

Thanks to point me about `apt.conf.d/docker-clean` dark magic

PR #77, opened 2017-09-26
Adding ros1 source to entrypoint for ros2_bridge

ruffsl:


ruffsl:
@mikaelarguedas , I made the entrypoint for the ros bridge separate. I there still that general python issue with the colliding ros1/ros2 msg paths, or could I just put these changes in the main entrypoint.

youtalk:
@ruffsl I think someone confuses the naming `ros2_bridge_entrypoint.sh` Why don't you rename to `ros1_bridge_entrypoint.sh` instead?

mikaelarguedas:
I agree with @youtalk the name is confusing (not clear if we refer to a port of the ROS 1  `ros_bridge` package, the `ros1_bridge` ROS 2  package or something else) so :+1:  for using `ros1_bridge_entrypoint.sh` as suggested by @youtalk.

The message packages names are still the same between ROS 1 and ROS 2, so their position in the `PYTHONPATH` still matters (the first one on the path is the one you get when you try to import)

ruffsl:
`ros2_bridge_entrypoint.sh` -> `ros1_bridge_entrypoint.sh`

youtalk:
@ruffsl Thank you.

> The message packages names are still the same between ROS 1 and ROS 2, so their position in the PYTHONPATH still matters (the first one on the path is the one you get when you try to import)

@mikaelarguedas It's true. I think [only sourcing `ros1_bridge/local_setup.bash`](https://github.com/osrf/docker_images/pull/76/files#diff-d1a8cbfb293d998111a48b38e16f1515R8) is less effective so far.

mikaelarguedas:
> I think only sourcing ros1_bridge/local_setup.bash is less effective so far

Yes, we need to source both because the bridge (C++) needs both to use `roscpp` and `rclcpp` (ROS 2) to create matching publishers and subscribers between ROS 1  and ROS 2. The collision happens only for Python message packages that are not used by the bridge.

youtalk:
@ruffsl lgtm too. Please merge this and rebuild `osrf/ros2:r2b3-ros1-bridge` image.

Issue #84, opened 2017-10-01
gzweb7 Build Fails: node-gyp error


Kukanani:
Hi,

I'm trying to build the gzweb7 docker, but it fails with the following output:

```
...
[100%] Building CXX object tools/CMakeFiles/gzcoarse.dir/gzcoarse.cc.o
Linking CXX executable gzcoarse
[100%] Built target gzcoarse
gyp info it worked if it ends with ok
gyp info using node-gyp@3.6.2
gyp info using node@0.10.25 | linux | x64

/root/gzweb/node_modules/node-gyp/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
const Hoek = require('hoek');
^^^^^
gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack SyntaxError: Use of const in strict mode.
gyp ERR! stack     at Module._compile (module.js:439:25)
gyp ERR! stack     at Object.Module._extensions..js (module.js:474:10)
gyp ERR! stack     at Module.load (module.js:356:32)
gyp ERR! stack     at Function.Module._load (module.js:312:12)
gyp ERR! stack     at Module.require (module.js:364:17)
gyp ERR! stack     at require (module.js:380:17)
gyp ERR! stack     at Object.<anonymous> (/root/gzweb/node_modules/node-gyp/node_modules/request/node_modules/hawk/lib/index.js:5:33)
gyp ERR! stack     at Module._compile (module.js:456:26)
gyp ERR! stack     at Object.Module._extensions..js (module.js:474:10)
gyp ERR! stack     at Module.load (module.js:356:32)
gyp ERR! System Linux 4.4.0-96-generic
gyp ERR! command "node" "/root/gzweb/node_modules/.bin/node-gyp" "configure"
gyp ERR! cwd /root/gzweb/gzbridge
gyp ERR! node -v v0.10.25
gyp ERR! node-gyp -v v3.6.2
gyp ERR! This is a bug in `node-gyp`.
gyp ERR! Try to update node-gyp and file an Issue if it does not help:
gyp ERR!     <https://github.com/nodejs/node-gyp/issues>
gyp info it worked if it ends with ok
gyp info using node-gyp@3.6.2
gyp info using node@0.10.25 | linux | x64
gyp ERR! build error 
gyp ERR! stack Error: You must run `node-gyp configure` first!
gyp ERR! stack     at /root/gzweb/node_modules/node-gyp/lib/build.js:49:20
[...]

PR #87, opened 2017-10-03
Update gzweb7 to use node v6.1.0

iche033:
see issue #84 

the `default` branch in gzweb no longer works with node v0.10 that comes with ubuntu trustry so updated the Dockerfile to install 6.1.0

### DIFF
--- a/gazebo/gazebo7/gzweb7/Dockerfile
+++ b/gazebo/gazebo7/gzweb7/Dockerfile
@@ -12,9 +12,6 @@ RUN apt-get update && apt-get install -q -y \
     libjansson-dev \
     libtinyxml-dev \
     mercurial \
-    nodejs \
-    nodejs-legacy \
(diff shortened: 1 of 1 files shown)


Kukanani:
Should the same changes be made for gzweb8? Or does it only target xenial, which doesn't have the node versioning issues? It would be nice to have consistent behavior and allow running gzweb8 in trusty, if it's a supported platform.

iche033:
These changes shouldn't be required for gzweb8 as it targets xenial which has a newer version of node

ruffsl:
Also, we may migrate gazebo 7 to xenial as well, see https://github.com/osrf/docker_images/issues/92

ruffsl:
The issue for this PR has been resolved: https://github.com/osrf/docker_images/issues/84
As gazebo7 dockerfiles now build from xenial.

PR #89, opened 2017-10-04
Migrating ROS2 to use new manifest

ruffsl:
No Dockerfile have changed in the migration,
running create_dockerfolder results in no further dockerfile changes as of writing.

### DIFF
--- a/ros2/manifest.yaml
+++ b/ros2/manifest.yaml
@@ -1,0 +1,40 @@
+%YAML 1.1
+# ROS2 Dockerfolder manifest
+---
+defaults:
+
+    default_hooks: &DEFAULT_HOOKS
+        hook_names:
+            post_push: .hooks/post_push.em
+
+    default: &DEFAULT
+        templates: &DEFAULT_TEMPLATES
+            images: .config/images.yaml.em
+            makefile: .config/Makefile.em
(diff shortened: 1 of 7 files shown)


ruffsl:
@tfoote or @mikaelarguedas 
All test are go, please review!

ruffsl:
Thanks, I'll merge at the end of the day to continue with upcoming CI PR automation

ros2/manifest.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+    default: &DEFAULT
+        templates: &DEFAULT_TEMPLATES
+            images: .config/images.yaml.em
+            makefile: .config/Makefile.em
+            platform: .config/platform.yaml.em
+
+release_names:
+    r2b3:
+        eol:
```
mikaelarguedas: Do we plan on adding eol dates after the fact ? beta 3 will be eol as soon as the next release is out (mid-december) and we will not support it from there.
slightly related question, when do we plan on removing the beta2 images ?

ruffsl: I don't plan we would include the pre-releases into the official library, so this manifest will be overwritten when the stable release is made. Its included currently just to conform to CI scaffolding.

Issue #53, opened 2017-05-24
Adding CI support for health monitoring of Dockerfile builds


ruffsl:
### Issue
Currently this repo has no internal continuous integration, thus maintainers are given little notice of build issues unless attempting to rebuild locally; typically via someone noticing lapse of refreshing of images on Dockerhub even as upstream images are updated or via official library maintainer feedback during periodic PRs while adding new changes. I'd like to recommend adding an amount of continuous integration for monitoring build health, for at least our official library repos/tags and perhaps for local PRs as well. 

### Context
This need arises from an issue during the latest release cycle where a subtle change in a upstream parent image, a sudden packages omission in ubuntu, silently resulted in build failures of the official ROS Dockerhub images for a considerable about of time, leaving our official images stale and vulnerable from lack of updates. Specifically, the debian package `locals` was previously required to generate an expected charmap in setting the system locale. This has been mitigated by moving setup specific commands after the building image has been primed with appropriate `apt` installs in previous layers. However one should prepare for the unexpected, thus notifications of such occurrence would help improve our response time.

Additionally, CI support of PR validation would also be beneficial, e.g. checking all modified Dockerfiles in a PR continue to build. This would be helpful as a maintainers are not always in possession of suitable network bandwidth to test build Dockerfiles locally, i.e. using ` --no-cache`. This occurred while I was attempting to fix `locals` issue for past distro tags, my ROS generation script prematurely bumped up the ros meta-package version numbers. Specifically the version info scraped from the rosdistro repo was ahead of what was at the time synced into OSRF's package repo, thus resulting in a build failure I didn't anticipate.  

### Proposal
I'd like to propose using some CI service such as travis or the like to trigger builds when say either a respective parent image is pushed to DockerHub's registry, or when a sync to a respective distro/version is released onto the package repo. For good measure, a weekly periodic build may also be useful to spot intermittent issues. Ideally I would've liked to tie in directly to whatever the official library managers use to monitor images, but it doesn't seem such exists yet or isn't exposed externally. 

We could use webhooks from DockerHub's registry to sense pushes to upstream parent images, and perhaps something equivalent can be done syncs to OSRF's package repo. Github integration for PRs shouldn't be too tricky, although adjusting it to focus on only affected tags may require some more machinery. Another tricky aspect is that due to the nesting of tags our images use to optimize image layer reuse, build order of tags while testing will need to respect the order of parent image hierarchy. If folks have ideas or example in testing dockerfiles using CI, please voice it here.

[...]

PR #94, opened 2017-10-06
Adding auto generated PR support for CI

ruffsl:
This fixes https://github.com/osrf/docker_images/issues/53.

Currently this is intended to enable Travis to trigger pull requests via a github machine account using personal tokens. The machine user does not need push access to the public repo, as it simple submits PR from the head of branches pushed to its own fork of the repo.

For cron jobs testing the default branch, it first checks that no changes occur in the dockerfile auto generation. If none are found, then it proceeds to test build the Dockerfiles via Docker. If changes have occurred, then it commits the changes, updates and commits the docker library manifest, then pushes it to a relatively named branch, and makes the PR. If the branch already exist, it force pushes to the branch, automatically updating any PRs that could still be open, keeping them from going stale. The PR CI then dogfood tests that those file changes do in fact build with Docker.


ruffsl:
One issues I've had so far is that everything works if I run the script locally, but when I run it from travis, it seems that something is somehow different, and creating the PR fails; 

#### Local
https://github.com/ruffsl/docker_images/pull/19
```
export HUB_REPO='ros'
export HUB_RELEASE='lunar'
export HUB_OS_NAME='ubuntu'
export HUB_OS_CODE_NAME='xenial'

export TRAVIS_BRANCH='master'
export TRAVIS_PULL_REQUEST_BRANCH=''
export TRAVIS_REPO_SLUG='ruffsl/docker_images'
export TRAVIS_BUILD_DIR='/home/ruffsl/Downloads/git/ruffsl/docker_images'

export GITHUB_USER='<>'
export GITHUB_EMAIL='<>'
export GITHUB_TOKEN='<>'
.travis/travis.py
...
tags:  ['lunar-perception-stretch']
Traceback (most recent call last):
  File ".travis/travis.py", line 141, in main
    test_diffs(diffs)
  File ".travis/travis.py", line 49, in test_diffs
    raise ValueError('Autogenerated files are not up to date')
ValueError: Autogenerated files are not up to date

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".travis/travis.py", line 246, in <module>
    main()
  File ".travis/travis.py", line 238, in main
    "has been created.").format(pr_head_name=pr_head_name))
ValueError: Relevent PR from osrf-docker-builder:ros/lunar/ubuntu/xenial has been created.
```
#### Travis
https://travis-ci.org/ruffsl/docker_images/jobs/284051612
```
tags:  ['lunar-perception-stretch']
Traceback (most recent call last):
  File ".travis/travis.py", line 140, in main
    test_diffs(diffs)
  File ".travis/travis.py", line 49, in test_diffs
    raise ValueError('Autogenerated files are not up to date')
ValueError: Autogenerated files are not up to date
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File ".travis/travis.py", line 244, in <module>
    main()
  File ".travis/travis.py", line 233, in main
    head=pr_head_name)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/github/Repository.py", line 994, in create_pull
    return self.__create_pull_1(*args, **kwds)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/github/Repository.py", line 1003, in __create_pull_1
    return self.__create_pull(title=title, body=body, base=base, head=head)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/github/Repository.py", line 1016, in __create_pull
    input=post_parameters
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/github/Requester.py", line 172, in requestJsonAndCheck
    return self.__check(*self.requestJson(verb, url, parameters, headers, input, cnx))
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/github/Requester.py", line 180, in __check
    raise self.__createException(status, responseHeaders, output)
github.GithubException.GithubException: 422 {'message': 'Validation Failed', 'errors': [{'resource': 'PullRequest', 'code': 'custom', 'message': 'No commits between ruffsl:master and [secure]:ros/lunar/ubuntu/xenial'}], 'documentation_url': 'https://developer.github.com/v3/pulls/#create-a-pull-request'}
The command ".travis/travis.py" exited with 1.
Done. Your build exited with 1.
```

Perhaps @nuclearsandwich might know what I'm missing?

ruffsl:
Ok, I found that last issue. I forgot travis create a detached head as opposed checking into the origin branch (like I had tested locally). So instead of try to push master, that incurred no commit changes, I just create a new named branch from the detached head and push that instead.

Check out the awesome joke the bot used in the example PR: https://github.com/ruffsl/docker_images/pull/23

nuclearsandwich:
> Perhaps @nuclearsandwich might know what I'm missing?

It looks like you've got some commits to try and resolve the issue. I'm still not quite adept at reading python stacktraces.  The creation of the pull request failed because there were "No commits between ruffsl:master and [secure]:ros/lunar/ubuntu/xenial". So either the wrong local branch was pushed to the upstream branch or the local branch was created at the wrong time.

I will again suggest replacing the use of a branch,add,commit machine workflow with a commit-tree,update-ref workflow. As doing so dodges a lot of hairy repository state related problems that are difficult to debug.

An example of commit-tree use in a similar automated commit context is [here](https://github.com/gjtorikian/repository-sync/blob/164b02b5d81d8d01c0ca26bd149a93d1cab69637/lib/cloner.rb#L236-L237).

mikaelarguedas:
> Check out the awesome joke the bot used in the example PR: ruffsl#23

Haha, great joke, I was waiting for this glorious day where I wouldn't have to write tests anymore.

I'll try to review this soon

.travis/travis.py
```diff
@@ -16,95 +48,199 @@ def test_diffs(diffs):
...
+    GIT_AUTHOR = "{user} <{email}>".format(user=GIT_USER, email=GIT_EMAIL)
+    GIT_ORIGIN_REPO_SLUG = GIT_USER + '/' + GIT_UPSTREAM_REPO_SLUG.split('/')[1]
+    GIT_REMOTE_ORIGIN_TOKEN_URL = "https://{user}:{token}@github.com/{repo_slug}.git".format(
+        user=GIT_USER,
+        token=GIT_TOKEN,
+        repo_slug=GIT_ORIGIN_REPO_SLUG
+    )
+
+    # Initlise git interfaces
+    repo = git.Repo(GIT_BUILD_DIR)  # , odbt=git.GitCmdObjectDB)
```
mikaelarguedas: Is the goal to provide an object database in the future? or can this comment be removed?

.travis/travis.py
```diff
@@ -16,95 +48,199 @@ def test_diffs(diffs):
...
+                    '--output', output))
+
+                # Commit changes to Docker Library
+                repo.git.add(all=True)
+                message = "Updating Docker Library\n" + \
+                    "This is an automated CI commit"
+                repo.git.commit(m=message, author=GIT_AUTHOR)
+
+                # Create new branch from current head
+                new_branch = repo.create_head(pr_branch_name)
```
mikaelarguedas: I guess that this was necessary to fix the problem mentionned in https://github.com/osrf/docker_images/pull/94#issuecomment-334830221.
What I'm surprised to see is that this branch variable is not used anywhere in the code (same thing for g_branch below). Am I missing something here?

.travis/travis.py
```diff
@@ -16,95 +49,198 @@ def test_diffs(diffs):
...
+                    # Continue if branch does not already exist
+                    if exception.data['message'] == "Branch not found":
+                        try:
+                            repo.git.push(
+                                'upstream_pr', pr_branch_name + ':' + pr_branch_name, u=True)
+                        except Exception as inst:
+                            inst.stderr = None
+                            raise ValueError(
+                                ("Set-upstream push to branch:{branch} failed! "
+                                 "Stderr ommited to protect secrits.").format(branch=pr_branch_name))
```
mikaelarguedas: what happens if the error message is NOT "Branch not found", I think it should raise the exception instead of continuing isn't it?

.travis/travis.py
ruffsl: I just wanted to capture what they returned so it would be clear that those function return specific object types, incase I have to re-read the code later. Would it be best to omit the variables?

.travis/travis.py
ruffsl: I wasn't sure if I should should use [GitCmdObjectDB](http://gitpython.readthedocs.io/en/stable/tutorial.html#gitcmdobjectdb) or stick with [GitDB](http://gitpython.readthedocs.io/en/stable/tutorial.html#gitdb). I'll try timing it, and use whatever is better.

.travis/travis.py
mikaelarguedas: makes sense, in that case you can add "  # noqa" at the end of the line to make linters/test editors happy

.travis/travis.py
mikaelarguedas: sounds good

.travis/travis.py
ruffsl: Fixed

ruffsl:
@mikaelarguedas I simplified the puch logic.

mikaelarguedas:
thanks @ruffsl for iterating on this, I pushed a few trivial fixes (style/typos) but the code looks good!

Issue #93, opened 2017-10-05
Adding tags for ROS releases built from non LTS Linux distros


ruffsl:
As the title suggest, should we add such new tags?

Given the current auto generating scripts and CI, plus the upcoming PR bot, adding then maintaining such tags should be of minimal effort. This would provide users and developers with ROS installations on top of the latest OS distros, helping to improve early development and migration of packages, This would also eventually be useful for transitioning and adoption for the next eventual LTS target beyond. :+1: 

My one suggestion going forward would be to mandate that OS agnostic tags consistently point to images built from LTS OS distros, and that the `latest` points to the LTS ROS release built from a LTS OS distro.

An alternative to hosting the full matrix of OS and ROS tags, somthing that could end up being a little overwhelming :dizzy: either for users or mainters, we could just make it so non LTS ROS releases are just build simply from the latest OS supported, even if it's not and LTS. But given the current OS related tags already listed for ROS, the full matrix approach would avoid removing any tags.

I think some of this might also touch on how we should approach EOL'ing tags: https://github.com/osrf/docker_images/issues/79

tfoote:
If were're auto generating them adding them does seem reasonable. There's some slightly lower overhead to only doing the latest, but most only do one or two non-lts releases and keeping track of latest seems like more metadata to be maintained, and the cleanup doesn't sound like something that would be good to need to do.

mikaelarguedas:
> My one suggestion going forward would be to mandate that OS agnostic tags consistently point to images built from LTS OS distros, and that the latest points to the LTS ROS release built from a LTS OS distro.

:+1: 

> but most only do one or two non-lts releases and keeping track of latest seems like more metadata to be maintained

I agree, if providing them all is low overhead let's do that.

> I think some of this might also touch on how we should approach EOL'ing tags: #79

For EOL tag I do still strongly believe that as soon as a distro is EOL or a ROS distro is EOL we should stop updating and rebuilding the docker images. This way we can provide the closest image to what was last built on the buildfarm. Users can always update their system afterwards but we won't have to support any breakage that this could cause

ruffsl:
So what exactly would we like to add that is currently active and missing?
- lunar-ros-<>-yakkety ?
- lunar-ros-<>-zesty

mikaelarguedas:
I think that zesty is the highest value these days (close to what artful will be dependency-wise).
Yakkety I don't care much about it, given that it's already EOL it would be a one time build just to say we provide it but we won't make any updates to it

[...]

PR #96, opened 2017-10-07
 Adding Ubuntu zesty for ROS Lunar

ruffsl:
Relates to https://github.com/osrf/docker_images/issues/93


ruffsl:
Shoot, lunar+zesty is only for amd64 right?

PR #106, opened 2018-01-03
Remove unnecessary hack and set bashrc for SROS1 image

ruffsl:
For simplifying tutorials elsewhere.


nuclearsandwich:
What makes the hack unnecessary now? What made it necessary previously?

ruffsl:
This was me simplifying the experience for example image for tutorials in a sros paper. 

The bash hack was a carry over from elsewhere when I didn't know that the catkin build environment could be expected via argument rather that resorting to sourcing a bash file. Of which was never used to begin with, given this catkin invocation here, having no ros Deb installed to opt and this a complete source build.

PR #108, opened 2018-01-09
[ROS2]Ardent support

### LINKED ISSUE #107: [ROS2] Replace r2b3 with ardent
We should update the manifest and regenerate the docker files now that Ardent has been released.
We'll also need to update the source docker file and add the new dependencies (https://github.com/osrf/docker_images/compare/testing_ardent_binaries#diff-611c8dba035afebd27cbc82b2b788f49)

mikaelarguedas:
This removes r2b3 in favor of ardent and update the dockerfiles accorndingly.
Holding the merge waiting for a decision to be made on the tags

fixes #107


ruffsl:
@mikaelarguedas , I like those tags you've added. You can set the `version: True` in the `platform.yaml` to fixate the package versions. However, it looks like you've added wildcards to the package list in `images.yaml`, so the current docker_templates python module would be unable to resolve those.

Do you intend to keep the wildcards, or is this temporary? We could adapt the docker_templates library to accommodate this use case, i.e. expand the list from wildcards and then tag the version for each specific package. Or is the intent to alway pass the wildcard to apt directly to be inferred at build time? I think the later would be simpler for auditing purposes and for upstreaming, even if the dockerfiles themselves turnout to be longer.

mikaelarguedas:
@ruffsl Thanks for the feedback. 
TL;DR: I think having the expansion of wildcard and population of the dockerfiles with each package and its version would be great, but I don't think it's required for this release.

> Do you intend to keep the wildcards, or is this temporary? We could adapt the docker_templates library to accommodate this use case,

Yeah I was wondering the same thing. It's currently a convenience to be able to use the wildcards as these tags / variants are likely to change in the future (as I make them up as I go) and because we don't have "metapackages" associated with them that would stremline the installation process).

I don't think there is a need to pin down versions right now as we don't plan on submitting any of these upstream at the moment and that would allow us to trigger rebuilds for each ardent patch release without having to modify the dockerfiles. So I'd say it would be great in the future but it shouldnt be necessary to spend time on it right now.

PR #113, opened 2018-01-26
Ros2 entrypoint cleaning

ruffsl:
It seems specifying an entrypoint without a default command afterward results in the behavior where images do not necessarily inherent the base image's default command. This is a bit unitutve for users, as when no shell executable command is passed to the docker run, it does not drop into a shell session as it does only with the core image.

This also needs https://github.com/osrf/docker_templates/pull/31

### DIFF
--- a/ros2/.config/images.yaml.em
+++ b/ros2/.config/images.yaml.em
@@ -18,7 +18,6 @@ images:
         base_image: osrf/@(user_name):@(ros2distro_name)-core
         maintainer_name: @(maintainer_name)
         template_name: docker_images/create_ros2_image.Dockerfile.em
-        entrypoint_name: docker_images/ros2_entrypoint.sh
         template_packages:
             - docker_templates
         packages:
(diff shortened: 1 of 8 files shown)


ruffsl:
I noticed this from the wiki, and wondered why I needed to declare bash as my command for the basic image here: https://github.com/ros2/ros2/wiki/Run-2-nodes-in-a-single-docker-container

mikaelarguedas:
Yeah I noticed that to and wanted to fix it :+1: . One thing I wanted to do as well is to source the bash file for autocompletion (likely in the `basic` image as that is the one introducing the command line tools). We can either do it in a follow PR or as part of this pair of PRs. As you prefer

ruffsl:
@mikaelarguedas , do you mean something like this:
https://github.com/keymint/keymint_tools/blob/master/.docker/keymint_tools/Dockerfile#L21

mikaelarguedas:
yup, 
(newbie question) What is the advantage of putting that specific source command in the bashrc rather than in the entrypoint (where we source the workspaces)?

tfoote:
Yeah, I'd generally avoid setting things into .bashrc automatically or by default. That should be the users choice not required on them.

Issue #116, opened 2018-01-31
Base Image for Gazebo 9


ruffsl:
From [Gazebo website](http://gazebosim.org/), gazebo v9 is supported until 2023, I assume in sync for [Ubuntu EOL ](https://wiki.ubuntu.com/Releases) of 18.04. However, as Bionic is yet to be released, and as v9 currently targets 16.04 and 17.10, the question is how we should proceed currently. 

My suggestion would be be that we would target 16.04 for now, with added tags with a `-xenial` postfix, then switch to bionic for the basic tags to 18.04 when supported, while preserving the `-xenial` tags until 16.04's EOL. Basically what we already do with ROS, but now ported for gazebo.

j-rivero:
Thanks @ruffsl, sounds good to me. Let me know if you need me to proceed with any code change.

Issue #112, opened 2018-01-26
[ros] Provide exact version numbers to ensure rebuilds


mikaelarguedas:
Context in https://github.com/docker-library/official-images/issues/3890.

TL;DR the docker images don't get rebuilt as the version of the metapackages never changes. We need a way to get them re-triggered.
 

Possible solution: provide exact version numbers for the packages in docker files so that the docker cache is invalidated. One con is that these version differ for each platform / arch so we need to find a way to conditionally set the right version for each arch without duplicating dockerfiles.

ruffsl:
What do you suppose would be our options here, in addition to what tianon proposed in  https://github.com/docker-library/official-images/issues/3890 ? It would be a shame that updating one single arch would consequently brake the cache of all the others. Not sure how avoid that given a single Dockerfile and different version strings. @nuclearsandwich, Do you spot a different angle?

nuclearsandwich:
What if we added a property to the Dockerfile that was an environment variable with a sync timestamp. Similar to the "daily invalidation" datestamp we use to conditionally rebuild docker images on the ROS buildfarm.

If I'm not mistaken that would be a change sufficient to re-trigger image generation, it's architecture independent, easy enough to do by hand and it's possible to automatically generate the PR in the future.

mikaelarguedas:
> What if we added a property to the Dockerfile that was an environment variable with a sync timestamp. Similar to the "daily invalidation" datestamp we use to conditionally rebuild docker images on the ROS buildfarm.

Yeah that's be the approach I'd take as well :+1:

ruffsl:
> added a property to the Dockerfile
Can you link to the example? Do you mean a dockerfile build arg?

mikaelarguedas:
In ros_buildfarm and ros2 we invalidate dockerfiles on a regular basis.
Some of them are invalidated daily. This is done by adding an echo statement in the Dockerfile at empy expansion time: https://github.com/ros-infrastructure/ros_buildfarm/blob/2499ffaf66e311bfd4710ac3798bf7fbef051a3d/ros_buildfarm/templates/doc/doc_independent_task.Dockerfile.em#L31

What Steven! is suggesting is that we expand the templates of a given distro after each sync / print the sync date in it and submit the new hash upstream.

Ideall this would be done in a job triggered automatically after each sync.

ruffsl:
Ok, so it wouldn't avoid breaking cache of other arch pointing to the same dockerfile,  
[...]

Issue #79, opened 2017-09-26
Handling of EOL ROS distros


ruffsl:
As ROS distos enter end of life, there is the question as to how we should maintain their presence on official library/ros repo. As observed, the current practice seems that once a version release under a given tag reaches EOL, the tag is eventually omitted from the build listing for the library, and new images for the tag are no longer pushed to the registry given the release will no longer receive updates.

Such is the case I see for ubuntu, where older images for 10.04 or 12.04 present within the registry and can be pulled by users, though legacy tags are not blatantly advertised on the doc pages, encouraging users to migrate to supported releases, legacy tags are also not deleted from the registry, preserving the archive.

Given that:
 * Indigo and later have nested tag that are not replicable without the official library, and
 * These distros where at one time supported by the official library

My own opinion would be that these tags should be properly archived in library/ros as well. More specifically, I think that once the upstream base image for a tag in library/ros is no longer updated by the docker hub library, only then should we really remove ROS tags from the official build manafest.

This touches another issue; given non LTS ROS releases tags, such as Jade, targeting LTS base images, such as Ubuntu 14.04, be left active in the build manifest for the duration of the base image support?

One may argue:
 * Removing a tag prematurely before the final push of its base image would forever leave the tag out of sync with upstream, thus lacking any final security updates or system patches. 
 * Alternatively, halting the pushing of the tag at a known state could help prevent regressions between the duration of the EOL of the tag, and final base image push.

I think the second point would be mitigated by any such changes breaking the build at the deb install process, given the pinned version numbers embedded in the dockerfile preventing any consecutive pushes from regressing the tags before it is omitted from the build manafest. But perhaps there are some unforeseen runtime variability I am unaware of.

What I think this really comes down to is what state we wish to leave images once they reach EOL. Would we prefer to leave the tagged images at up to date as we can, and drop support with the finality of the EOL of the ROS distro? Or try and generate the operability of tag after it is omitted from the build manafest?

pringing: @mikaelarguedas @tfoote @j-rivero @tianon @yosifkit

yosifkit:
For any image @tianon and I directly maintain, we drop a version as soon as it is EOL; there are simply too many we maintain to keep them longer than is absolutely necessary.

If you want to keep the EOL ROS releases and are willing to continue maintaining the image and ensuring they continue to build, then I don't have any objections. What do you think @tianon?

[...]

PR #117, opened 2018-01-31
Adding Gazebo9 for xenial

ruffsl:
Fixes https://github.com/osrf/docker_images/issues/116

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
   - HUB_REPO=ros    HUB_RELEASE=kinetic HUB_OS_NAME=debian HUB_OS_CODE_NAME=jessie
   - HUB_REPO=ros    HUB_RELEASE=indigo  HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=trusty
   # gazebo
+  - HUB_REPO=gazebo HUB_RELEASE=9       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
   - HUB_REPO=gazebo HUB_RELEASE=8       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
   - HUB_REPO=gazebo HUB_RELEASE=7       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
   - HUB_REPO=gazebo HUB_RELEASE=6       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=trusty
--- a/gazebo/9/ubuntu/xenial/Makefile
+++ b/gazebo/9/ubuntu/xenial/Makefile
@@ -1,0 +1,28 @@
+all: help
+
+help:
+	@echo ""
+	@echo "-- Help Menu"
+	@echo ""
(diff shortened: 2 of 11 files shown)


ruffsl:
Let me know if 5c09518 is worth including.

mikaelarguedas:
> Let me know if 5c09518 is worth including.

I don't think it's needed but it cannot hurt, especially for the not-EOL ones that are targetting multiple distros.
Regarding the EOL ones: Does it make sense to update and rebuild them ? (IDK if there has been discussion similar to https://github.com/osrf/docker_images/issues/79 for gazebo images)

ruffsl:
> Regarding the EOL ones: Does it make sense to update and rebuild them ?

Those Dockerfiles have not changed, so the docker build farm should just add the tag to the catched images. However, if we modified all templates to include the [complete version](https://github.com/osrf/docker_images/issues/112#issuecomment-361806162), that would trigger a rebuild. Before that though, we could just comment them out from the manifest like we did with Jade to avoid that.

mikaelarguedas:
> Those Dockerfiles have not changed, so the docker build farm should just add the tag to the catched images. 

I meant rebuilding them for each base image change.

> if we modified all templates to include the complete version, that would trigger a rebuild

Yeah I'm still not convinced this approach brings us much benefits compared to the one suggested by the docker maintainers. But let's keep this discussion separated for now.

> Before that though, we could just comment them out from the manifest like we did with Jade to avoid that.

:+1: yup, that sounds good to me (relevant PR for reference: #109), we should check with the gazebo team though.
If you prefer I can open a separate issue to discuss EOL gazebo images if that allows us to get this one merged in a timely manner

ruffsl:
> I meant rebuilding them for each base image change.

I don't understand, none of the EOL gazebo tag are changing base images here. Do you mean when eventually bumping the plane `libgazebo` or `latest` tags when targeting the latest supported base LTS image when released?

> If you prefer I can open a separate issue to discuss EOL gazebo images if that allows us to get this one merged in a timely manner

Sounds good.

mikaelarguedas:
> I don't understand, none of the EOL gazebo tag are changing base images here. Do you mean when eventually bumping the plane libgazebo or latest tags when targeting the latest supported base LTS image when released?

Sorry that was unclear: as ubuntu images are the lowest level base image for all gazebo images, they all get rebuilt when a new ubuntu image comes out, right ?

mikaelarguedas:
@ruffsl opened #118 to track the EOL distros discussion

ruffsl:
> as ubuntu images are the lowest level base image for all gazebo images, they all get rebuilt when a new ubuntu image comes out, right ?

Ah yes, true indeed. Though I suspect for official library image pipeline, cache breaks are fine grained to individual tag changes, rather than merely repo wide triggers as with automated docker hub user images. So (my guess), should amd64/ubuntu:xenial update, a build for arm32v7/ros:indigo would not be triggered due to it being dependant on a separate ubuntu tag and arch.

ruffsl:
Just tried it out with gzweb, all seems to work fine.

PR #128, opened 2018-04-16
ROS 2 Bouncy: Bionic + colcon source dockefile

mikaelarguedas:
WIP: this is experimental to prepare the bouncy release

main changes:
- target bionic
- use new build tool

### DIFF
--- a/ros2/source/source/Dockerfile
+++ b/ros2/source/source/Dockerfile
@@ -13,6 +13,9 @@ RUN . /etc/os-release \
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
+RUN apt-get update && apt-get install -q -y tzdata \
+    && rm -rf /var/lib/apt/lists/*
+
 # install packages
 RUN apt-get update && apt-get install -q -y \
     bash-completion \
--- a/ros2/source/source/Dockerfile
+++ b/ros2/source/source/Dockerfile
@@ -25,6 +25,7 @@ RUN apt-get update && apt-get install -q -y \
     cppcheck \
     git \
     libasio-dev \
+    libassimp-dev \
     libcurl4-openssl-dev \
     libeigen3-dev \
     libgles2-mesa-dev \
--- a/ros2/source/source/Dockerfile
+++ b/ros2/source/source/Dockerfile
@@ -63,6 +63,7 @@ RUN apt-get update && apt-get install -q -y \
 # install python packages
 RUN pip3 install -U \
     argcomplete \
+    catkin_pkg \
     flake8 \
     flake8-blind-except \
     flake8-builtins \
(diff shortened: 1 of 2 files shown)


ros2/source/source/Dockerfile
```diff
@@ -73,26 +78,37 @@ RUN pip3 install -U \
...
-    --cmake-args -DSECURITY=ON -- \
-    --isolated \
-    --parallel \
-    --symlink-install
+    --cmake-args " -DSECURITY=ON" " -DBUILD_TESTING=ON" \
+    --symlink-install \
+    --event-handler console_cohesion+
 
 # setup bashrc
+ENV ROS2_WS /root/ros2/ros2_ws
```
ruffsl: This is redundant given line 96

mikaelarguedas: You're totally right.
Sorry for wasting your review time, this is currently playground and far from being ready for review. I just pushed a rebased version because someone else needed to do some debugging on bionic.

I will open a PR once this branch is closer to a reviewable state

mikaelarguedas: If you prefer I can do my playground on a fork in the future to avoid spurious email notifications ?

ros2/source/source/Dockerfile
```diff
@@ -52,16 +55,18 @@ RUN apt-get update && apt-get install -q -y \
     python3-pip \
     python3-pyparsing \
     python3-setuptools \
-    python3-vcstool \
+    # python3-vcstool \
     python3-yaml \
     qtbase5-dev \
     uncrustify \
     wget \
+    xorg-dev \
```
sloretz: What needs `xorg-dev`? Is this for `rviz_ogre_vendor`? It built fine for me after only installing `libxrandr-dev`.

ros2/source/source/Dockerfile
```diff
@@ -73,26 +78,37 @@ RUN pip3 install -U \
...
+    vcstool
+
+# clone and build buildtool
+ENV COLCON_WS /root/ros2/colcon
+RUN mkdir -p $COLCON_WS/src
+WORKDIR $COLCON_WS
+RUN wget https://raw.githubusercontent.com/colcon/colcon.readthedocs.org/master/colcon.repos && \
+    vcs import src < colcon.repos
+RUN ./src/colcon-core/bin/colcon build --paths src/*
+RUN . install/prefix.sh && colcon build
```
sloretz: Master is now `setup.sh` instead of `prefix.sh`. I don't think @dirk-thomas has done a release with that change yet, but when he does how about using `pip` instead of bootstraping?

ros2/source/source/Dockerfile
```diff
@@ -1,18 +1,21 @@
...
-    && echo "deb http://repo.ros2.org/$ID/main $VERSION_CODENAME main" > /etc/apt/sources.list.d/ros2-latest.list
-
+# RUN . /etc/os-release \
+#     && echo "deb http://repo.ros2.org/$ID/main $VERSION_CODENAME main" > /etc/apt/sources.list.d/ros2-latest.list
+ENV DEBIAN_FRONTEND noninteractive
 # setup environment
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
+RUN apt-get update && apt-get install -q -y tzdata \
```
sloretz: Do any packages besides `tzdata` ask for user input when they're being configured? If not I think the `ENV DEBIAN_FRONTEND noninteractive` line above could be removed by making this line `RUN DEBIAN_FRONTEND=noninteractive apt-get update && ...`

ros2/source/source/ros2_source_entrypoint.sh
```diff
@@ -2,5 +2,5 @@
 set -e
 
 # setup ros2 environment
-source "$ROS2_WS/install_isolated/local_setup.bash"
+source "$ROS2_WS/install/prefix.bash"
```
sloretz: `setup.bash` here too

ros2/source/source/Dockerfile
ruffsl: Especially as currently dockerfile environment variables [can't just be unset](https://github.com/moby/moby/issues/3465) later by end users, and passing them inline seems to be a common [compromise](https://github.com/moby/moby/issues/9298#issuecomment-133750740).

ros2/source/source/Dockerfile
mikaelarguedas: yes I have these changes locally but haven;t pushed them yet as this is not ready for review. Do you need an up-to-date working version pushed here ?

ros2/source/source/ros2_source_entrypoint.sh
mikaelarguedas: same comment as https://github.com/osrf/docker_images/pull/128#discussion_r186231310

ros2/source/source/Dockerfile
mikaelarguedas: > but when he does how about using pip instead of bootstraping?

As I use this for development, I bootstrap it for now because I almost always need features that are not yet released. We can definitely move to pip when this is more stable and thee is a one-line update path for users

ros2/source/source/Dockerfile
mikaelarguedas: Yeah totally, I also have that change locally but waited for the final state of the ros2/ci PR to be merged to push all the changes to that branch

ros2/source/source/Dockerfile
mikaelarguedas: Thanks for checking that. Yeah narrowing down what's needed to build and run rviz deps and tests is the last remaining item for the bionic migration. Once confirmed on the ros2/ci PR we can push the corresponding changes here

ros2/source/source/Dockerfile
mikaelarguedas: resolved in https://github.com/osrf/docker_templates/pull/34

mikaelarguedas:
closing in favor of https://github.com/osrf/docker_images/pull/160

PR #136, opened 2018-05-18
packages -> upstream_packages

mikaelarguedas:
minimal changes to support https://github.com/osrf/docker_templates/pull/37

### DIFF
--- a/ros2/source/images.yaml.em
+++ b/ros2/source/images.yaml.em
@@ -9,7 +9,7 @@ images:
         entrypoint_name: docker_images/ros2_source_entrypoint.sh
         template_packages:
             - docker_templates
-        packages:
+        upstream_packages:
             - bash-completion
             - build-essential
             - clang-format
@@ -29,7 +29,6 @@ images:
             - libqt5gui5
             - libqt5opengl5
             - libqt5widgets5
-            - libopensplice67
             - libssl-dev
             - libtinyxml-dev
             - libtinyxml2-dev
@@ -47,11 +46,13 @@ images:
             - python3-pip
             - python3-pyparsing
             - python3-setuptools
-            - python3-vcstool
             - python3-yaml
             - qtbase5-dev
             - uncrustify
             - wget
+        packages:
+            - libopensplice67
+            - python3-vcstool
         pip3_install:
             - argcomplete
             - flake8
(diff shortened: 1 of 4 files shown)


ros2/source/images.yaml.em
```diff
@@ -47,11 +46,13 @@ images:
             - python3-pip
             - python3-pyparsing
             - python3-setuptools
-            - python3-vcstool
             - python3-yaml
             - qtbase5-dev
             - uncrustify
             - wget
+        packages:
```
ruffsl: Just curious in this distinction between `upstream_packages` and `packages`?  
So `packages` should now be considered as non-ros/gazebo package that are installable after the new sources are added?

mikaelarguedas: Yeah, I was wondering how to outline the distinction. Neither of these packages are ROS packages (they are not associated with a distribution or have a package.xml or relying on ROS build systems).
I considered [`ros_hosted_packages`](https://github.com/osrf/docker_templates/pull/37/commits/d17b6b7ffd8130cbebdaf2d5b09dddba10d9ee29#diff-42e639a543ecddda58a7fcf95a8b6641R45) but am open to suggestions on the naming

ruffsl: Perhaps we could name them ‘upstream_packages’ and ‘downsteam_packages‘ too differentiate the expected sources of the package.

mikaelarguedas: I considered it as well but `downstream_packages` is not necessarily semantically correct. In this example, `libopensplice67` and `python3-vcstool` are "dependencies" of the ROS2 packages that will be built after, but they happen to be hosted on the ros apt repositories.

So I didn't change it just now. If you prefer we can switch to "downstream_packages" for now and iterate on it if/when we come up with a better name ?

ruffsl: `midstream_packages` ?

> If you prefer we can switch to "downstream_packages" for now and iterate on it if/when we come up with a better name ?

For refactoring, a more descriptive name now, rather than vaguely  `packages`, would make it easier to iterate later.

mikaelarguedas: I'll go for `ros2_repo_packages` in order to get this merged and have an expressive name. We can then easily replace it in both the templates repo and here

PR #160, opened 2018-05-29
[ROS2] Bouncy source dockerfile using new templates, colcon and updated dependency list

mikaelarguedas:
Replaces #128 

This changes the behavior to build `master` instead of `release-latest` we `~should likely go back to `release-latest` before merging.~~ will come back and go back to use release latest once bouncy is tagged and released

Update the ros2 source Dockerfile to build bouncy on bionic.
This now uses colcon for the build tool and depends on https://github.com/osrf/docker_templates/pull/39


ros2/source/images.yaml.em
```diff
@@ -65,16 +84,15 @@ images:
...
         ws: /root/ros2_ws
-        ament_args:
+        colcon_args:
             - build
-            - --build-tests
             - --cmake-args
-                -DSECURITY=ON --
-            - --isolated
-            - --parallel
+                " -DBUILD_TESTING=ON" " -DSECURITY=ON" " --no-warn-unused-cli"
```
dirk-thomas: The leading spaces and quotes can be removed now.

Same below.

mikaelarguedas: Sweet :+1:

mikaelarguedas:
Placing this in review.
I'll open a follow up at Bouncy release time to replace master with release latest.
This way we keep in sync with the templates that are updated to use colcon

ros2/source/images.yaml.em
```diff
@@ -65,16 +66,15 @@ images:
...
         ws: /root/ros2_ws
-        ament_args:
+        colcon_args:
             - build
-            - --build-tests
             - --cmake-args
-                -DSECURITY=ON --
-            - --isolated
-            - --parallel
+                -DBUILD_TESTING=ON -DSECURITY=ON --no-warn-unused-cli
```
dirk-thomas: `BUILD_TESTING` should be `ON` by default.

Same below.

PR #165, opened 2018-06-28
Adding nvidia-docker2 env to desktop

ruffsl:
libglvnd0 seems to already be installed from ros-robot, so I just needed to set the ENVs.
Ties in with https://github.com/osrf/docker_templates/pull/41

One odd thing however is that I can not run rviz on melodic from a ubuntu 16.04 host running nvidia-docker2. Not sure why yet. Example:

<details>

``` bash
#!/usr/bin/env bash

# Runs a docker container with the image created by build_demo.bash
# Requires
#   docker
#   nvidia-docker2
#   an X server
# Recommended
#   A joystick mounted to /dev/input/js0 or /dev/input/js1


# Make sure processes in the container can connect to the x server
# Necessary so gazebo can create a context for OpenGL rendering (even headless)
XAUTH=/tmp/.docker.xauth
if [ ! -f $XAUTH ]
then
    xauth_list=$(xauth nlist :0 | sed -e 's/^..../ffff/')
    if [ ! -z "$xauth_list" ]
    then
        echo $xauth_list | xauth -f $XAUTH nmerge -
    else
        touch $XAUTH
    fi
    chmod a+r $XAUTH
fi

docker run -it --rm \
  --runtime=nvidia \
  --env DISPLAY \
  --env QT_X11_NO_MITSHM=1 \
  --env XAUTHORITY=$XAUTH \
  --volume "$XAUTH:$XAUTH" \
  --volume "/tmp/.X11-unix:/tmp/.X11-unix" \
  --volume "/etc/localtime:/etc/localtime:ro" \
  --volume "/dev/input:/dev/input" \
  --privileged \
  osrf/ros:melodic-desktop-bionic \
  bash -c "roscore & rviz"
```
<\details>

<details>

``` terminal

root@590ee20f885a:/# rviz
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-root'
[ INFO] [1530146048.886593550]: rviz version 1.13.1
[ INFO] [1530146048.886641829]: compiled against Qt version 5.9.5
[ INFO] [1530146048.886652215]: compiled against OGRE version 1.9.0 (Ghadamon)
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
[ INFO] [1530146050.064950979]: Stereo is NOT SUPPORTED
[ INFO] [1530146050.066417371]: OpenGl version: 4.5 (GLSL 4.5).
libpng warning: iCCP: known incorrect sRGB profile
Segmentation fault (core dumped)

```
<\details>


AustinDeric:
@ruffsl I also get the segfault.  I notice this image is not based on the nvidia/cuda ubuntu image therefore the `nvidia-smi command doesn't work.

I have some (and it looks like you do too) ros-nvidia images based on the  nvidia/cuda docker image. What not have that as the base layer here?  I believe that solves this problem.

ruffsl:
Ok, looks like it might be an issue with the current release of rviz:

> ros-melodic-rviz/now 1.13.1-0bionic.20180710.192735 amd64 [installed,local]

As I just created a child image that builds rviz from source using current devel branch and that seems to run fine without segfaulting.  See gist here:

https://gist.github.com/ruffsl/e7ca631a618ece3eb0be8e4bf168accb

ruffsl:
@tfoote @flx42 , I just checked this again today, and it seem the issue with rviz above has been resolved, and I can't reproduce the issue. Let me know if you have any suggestions.

ruffsl:
> Now that we have the ability to run docker images in nvidia docker without baking in the nvidia before hand

Sounds fair. Seeing how modifing the envs in images alone would still require additional run arguments to set the environment, display, drivers, etc, it makes sense to outsource such configuration directly to such tools as well.

PR #169, opened 2018-06-29
Updating gazebo 9 to bionic

ruffsl:
Correct me if mistaken: with Gazebo 9 support period (EOL 2023-01-25) matching that of Ubuntu Bionic, it would make sense to make the latest gazebo 9 tags eventually target Bionic. I have left the Xenial tags in place, and can be deprecated in due time with the EOL 16.04.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ env:
   - HUB_REPO=ros    HUB_RELEASE=kinetic HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
   - HUB_REPO=ros    HUB_RELEASE=indigo  HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=trusty
   # gazebo
+  - HUB_REPO=gazebo HUB_RELEASE=9       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=bionic
   - HUB_REPO=gazebo HUB_RELEASE=9       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
   - HUB_REPO=gazebo HUB_RELEASE=8       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
   - HUB_REPO=gazebo HUB_RELEASE=7       HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=xenial
--- a/gazebo/9/ubuntu/bionic/Makefile
+++ b/gazebo/9/ubuntu/bionic/Makefile
@@ -1,0 +1,28 @@
+all: help
+
+help:
+	@echo ""
+	@echo "-- Help Menu"
+	@echo ""
+	@echo "   1. make build            - build all images"
+	@echo "   2. make pull             - pull all images"
+	@echo "   3. make clean            - remove all images"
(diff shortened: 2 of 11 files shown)


mikaelarguedas:
:+1: gazebo 9 ship in ubuntu bionic by default and it definitely makes sense to provide gazebo9 images on that platform.
Chatting with @chapulina, in addition to keeping the xenial tags in place, it would be good to keep the short version of the tag mapping to Xenial for now. While Xenial is not the latest platform, it is the currently most commonly used and changing the base image from under users can have undesired side effects and may break many setups.
We can revisit what the "default" tags map to in the future, but I'd recommend not changing them in this PR and simply add bionic images.

ruffsl:
> While Xenial is not the latest platform, it is the currently most commonly used and changing the base image from under users can have undesired side effects and may break many setups.

Hmm, I think is would set a confusing precedent to have the `latest` tag no longer pointing to the latest LTS images, even if temporary, and would never expect any other official docker repos (e.g. Ubuntu) to diverge from such practices. Secondly, I think it would also unwise to encourage the user behavior of relying on the short tags to implectly remain tied to legacy distos, as that is instead the sole purpose of fully qualified tag names.  
I.e. `gazebo:libgazebo9-xenial`.

Short tags should be reserved for users who intentionally want their own image builds to follow the latest parrent image, e.g. for experimental CI to foreshadow migration issues when chasing upstream releases. Otherwise, users who do rely on the `latest` tag's intended use would have no other avenue to track upstream other than to manually watch official repos continuously for added tags.

I woudn't necessarily endorce that `latest` should point to most common, as it could 'check-or-egg' the adoption cycle, but how about we make a pre-announcement on the gazebo community form, [like last time](https://community.gazebosim.org/t/gazebo9-release/85/2), and give it some time (e.g. a week or more?) before merging and migrating the tags all in sync?

mikaelarguedas:
>  I think this would set a confusing precedent to have the latest tag no longer pointing to the latest LTS images, even if temporary, and would never expect any other official docker repos (e.g. Ubuntu) to diverge from such practices

Agreed

> but how about we make a pre-announcement on the gazebo community form, like last time, and give it some time (e.g. a week or more?) before merging and migrating the tags all in sync?

Yeah I would be fine with that as well.

I'll leave it up to @scpeters and @chapulina to pitch in on this before moving forward.

chapulina:
I'll defer to you peeps on Docker best practices. Everything @ruffsl said sounds reasonable to me.

mikaelarguedas:
@ruffsl feel free to merge and announce on discourse when appropriate.
As discussed above, we'll then let it soak for a few weeks before making the PR to `official_images`.

ruffsl:
> announce on discourse when appropriate.

I've cross posted to community.gazebosim.org, you think I should cross post to discourse.ros.org as well?
https://community.gazebosim.org/t/migrating-gazebo-9-docker-tags-to-bionic-base-image/209

mikaelarguedas:
> you think I should cross post to discourse.ros.org as well?

Yeah I think it doesn't hurt as there may be users following only one of the 2 platforms

mikaelarguedas:
Follow-up PR at https://github.com/osrf/docker_images/pull/185

PR #115, opened 2018-01-31
Version sync

ruffsl:
# WIP Do not merge
A sample PR for breaking every ROS tag into its own arch

PR #174, opened 2018-07-16
Add support for ARM64v8 architecture

odidev:
- Update manifest to fetch latest architecture based gazebo
   from available distributions.
   This will also work for AMD64 to fetch available packages.

This PR extends support for Gazebo package for ```arm64v8``` architecture.
A user needs to execute ```manifest.yaml``` to update code-base to generate all the available distribution packages.
Then ```docker build``` can be executed from the desired folder.

Currently,```arm64v8``` is supported in ```gazebo/9/debian/stretch/```

Signed-off-by: Odidev <odidev@puresoftware.com>


odidev:
Okay, finally **```All checks have passed```**.
Looks so good to see this message after so many conflict issues.

Guys, can you please check this **PR** and provide your valuable comments.

I have verified this on both ```amd64``` and ```arm64v8``` architectures.
I can also provide build and run logs as required for confirmation.

Regards,

ruffsl:
Looks like you added a duplicate entry for version 9. Instead you just need to insert `- arm64v8` under this line here: https://github.com/osrf/docker_images/blob/a136b533048de7eaeacd81db7707f3314cf7d297/gazebo/manifest.yaml#L115

and here:
https://github.com/osrf/docker_images/blob/a136b533048de7eaeacd81db7707f3314cf7d297/gazebo/manifest.yaml#L126

odidev:
Thanks @ruffsl 

I have modified this PR, as per your guidance.

Regards,

ruffsl:
Looks like Version '9.3.1-1*' is not currently available for arm64v8, as testing out the gazebo9 images with `FROM arm64v8/ubuntu:bionic` fails.

odidev:
@ruffsl 
Hi Ruffin,
Yes, I do agree with your concern.
Indeed ```9.3.1-1*``` is not present for ```arm64```.

Actually,```Gazebo9``` is not present for ```arm64``` from official gazebo url.
But it is there for ```debian```.

This is the available mapping right now at Gazebo URL for ARM64:
**Debian**
Stretch - Gazebo-9.0.0-2

**Ubuntu **
Xenial - Gazebo-8.0.0-1
Zesty   - Gazebo-8.0.0-2 

This was reason that my earlier commit had duplicate entries for Gazebo9, one for Debian and other for Ubuntu.

Can you please suggest what should be next approach.
Or can you take it at your end, as it will even more faster approach.

My target is to include ```gazebo``` in official list of ```arm64v8``` as soon as possible.

Regards,

ruffsl:
Presently, we pull the latest available release version by checking `packages.osrfoundation.org/gazebo`, e.g. [here](https://github.com/osrf/docker_templates/blob/19f42cc7f38a4df1f088dd6c314f113c971ac36c/docker_templates/packages.py#L30). Because only one dockerfile is used for all available architectures, this inevitably leads to issues when the latest release for other architectures are not in sync with amd64, the architecture used by default. This issue is being tracked here: https://github.com/osrf/docker_images/issues/112 .

Going forward, I'd expect we could transition pulling the yaml file that @mikaelarguedas linked to in [his comment](https://github.com/osrf/docker_images/issues/112#issuecomment-376716941), although I'm not sure gazebo has the same yaml artifacts from its build farm yet as well. Then we could adopt Either two different maintenance schemas, I.e. specifying exact version numbers on a per architecture basis (#112) or maintaining separate dockefiles per architecture (#115).

What evenly resolves #112 will most likely resolve this issue.

odidev:
Going by official support of Gazebo-9 for ARM64v8, I can see that this is supported for Debian:Stretch distribution instead of Ubuntu.

I have updated the PR to use Debian:Stretch.

Regards,

lag-linaro:
Why do we have to introduce a new distro (Debian) to support ARM64?

Ubuntu Bionic seems to work just fine.

PR #176, opened 2018-07-18
Ros2 bouncy variants

mikaelarguedas:
Make use of the ROS 2 variants.

Need to be considered before merging:
- ardent doesnt have such metapackages. We could consider backporting them but we'd need to audit dependencies to do so
- as bouncy and ardent now diverge in many aspects, it will be challenging to keep the infrastructure working out of the box for both. We could consider the default generation to be adapted to bouncy and make ardent require extra work as it's less likely to evolve

Note: images build fine locally (using the ROS2 testing apt repo), we'll have to wait until the packages are synced to the main repo for these dockerfile to build successfully


mikaelarguedas:
@ruffsl Headsup: I'm planning on merging this as is and revisit it if / when we provide variants for ardent.
If we dont, we will be able to go back to a single configuration files once Ardent goes EOL and Crystal gets released (as they will both have the variants)

ruffsl:
Looks good! BTW, was there a rep or decision on the use of meta packages in ros2? Last time I asked, this was still up in the air, but now looks settled. I'm just curious if such variants be can/can't be used as dependencies for other ament packages manifests?

mikaelarguedas:
There's not a set course of action so far.
Currently metapackages they are just regular ament packages (so no machinery to prevent them from having CMake code or being depended on). Just to clarify in ROS1 you can declare metapackages as runtime dependencies in the manifest of other packages, but (I think) you can't have particular CMake code with them or find_packaging them.

The original discussion you may be referring to is available at https://github.com/ros2/ros2/issues/408

As I couldn't find the time to explore more and / or implement the necessary additional parts, the current consensus was that we will keep them a simple ament packages for now and iterate later.

I opened a [REP Draft](https://github.com/ros-infrastructure/rep/pull/176) for the currently defined set though but it's specific to bouncy variants and doesnt cover the broader concept of metapackages.

mikaelarguedas:
I actually got around releasing these metapackages in ardent so I opened a simpler PR that update all dockerfiles to use them.

Closing in favor of #178

PR #177, opened 2018-07-19
use rosdep to install source dependencies

mikaelarguedas:
**WIP**
Maybe we provide the rosdep arguments in the images file and expand it in the template?

### DIFF
--- a/ros2/source/images.yaml.em
+++ b/ros2/source/images.yaml.em
@@ -11,48 +11,11 @@ images:
             - docker_templates
         upstream_packages:
             - bash-completion
-            - build-essential
-            - clang-format
-            - cmake
-            - cppcheck
             - git
             - libasio-dev
-            - libassimp-dev
-            - libcurl4-openssl-dev
-            - libeigen3-dev
-            - libfreetype6-dev
-            - libgles2-mesa-dev
-            - libglu1-mesa
-            - libopencv-dev
-            - libpoco-dev
-            - libqt5core5a
-            - libqt5gui5
-            - libqt5opengl5
-            - libqt5widgets5
-            - libssl-dev
-            - libtinyxml-dev
             - libtinyxml2-dev
-            - libxaw7-dev
-            - libxrandr-dev
-            - openssl
-            - pydocstyle
-            - pyflakes
-            - python3-coverage
-            - python3-dev
-            - python3-empy
-            - python3-mock
-            - python3-nose
-            - python3-pep8
-            - python3-pyparsing
-            - python3-pytest
-            - python3-pytest-cov
-            - python3-pytest-runner
-            - python3-setuptools
-            - python3-yaml
-            - qtbase5-dev
             - wget
         ros2_repo_packages:
-            - libopensplice67
             - python3-catkin-pkg-modules
             - python3-vcstool
         pip3_install:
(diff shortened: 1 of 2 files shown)


ruffsl:
> Maybe we provide the rosdep arguments in the images file and expand it in the template?

I added some commis to do this, relates to: https://github.com/osrf/docker_templates/pull/44

ros2/source/images.yaml.em
```diff
@@ -11,48 +11,13 @@ images:
...
-            - python3-pyparsing
-            - python3-pytest
-            - python3-pytest-cov
-            - python3-pytest-runner
-            - python3-setuptools
-            - python3-yaml
-            - qtbase5-dev
             - wget
+        downstream_packages:
+            - python-rosdep
```
mikaelarguedas: (I know it doesn't come from your changes but from my initial patch): As ROS 2 targets Python 3, can we install python**3**-rosdep here for consistency?

ruffsl: Fixed, also realized downstream template was redundant

mikaelarguedas:
LGTM once CI comes back green :+1: (can't approve my own PR :/)

ruffsl:
@mikaelarguedas , two more ideas:
1) Could we pip install everything, or is it better to use the release debs for the python tools, e.g.:
```
# install tools for ROS from pypi
RUN pip3 install -U \
    catkin_pkg_modules \
    colcon_common_extensions \
    rosdep \
    vcstool
```

2) Do we need to inti rosdep once with and without the new `ROSDISTRO_INDEX_URL`. Or is the default `ROSDISTRO_INDEX_URL` need to be init'ed at least once before?
``` diff
- RUN rosdep init \
-     && rosdep update
```

mikaelarguedas:
> Could we pip install everything, or is it better to use the release debs for the python tools, e.g.:

We usually try to avoid this as it means that user cannot update their environment by just running an apt upgrade (as the pip version takes priority). Meaning users need to also upgrade their pip packages to use the latest version of those packages.
As these packages are hosted on `repos.ros2.org` and that users will need these sources to install other packages like opensplice or connext I don't see a strong incentive for not installing them with apt.

> Do we need to inti rosdep once with and without the new ROSDISTRO_INDEX_URL. Or is the default ROSDISTRO_INDEX_URL need to be init'ed at least once before?

In this case it shouldn't matter as we are installing only system dependencies (so not impacted by the rosdistro we use, only the os version). So we could not set `ROSDISTRO_INDEX_URL` at all and it should work as well. But in theory we should point to the right rosdistro index if people happen to install released ros2 packages with rosdep. In that case the environment variable should be set *before* running `rosdep update` (doesn't matter if it is set before or after calling `init` I believe)

ruffsl:
> As these packages are hosted on repo.ros2.org and that users will need these sources to install other packages like opensplice or connext I don't see a strong incentive for not installing them with apt.

Good point about upgrading. I guess this Dockerfile could be used as minimal example for building from source, and would be good to follow best practices here.

mikaelarguedas:
> Merging as CI errors are unrelated to this PR.

haha nice timing I opened https://github.com/osrf/docker_images/pull/192 to be able to fix ci on this PR

Talking about CI: should we add the ros2 source to CI ? (it seems that it's currently not tested)

ruffsl:
> Talking about CI: should we add the ros2 source to CI ? (it seems that it's currently not tested)

If you think it wouldn't time out, I think it could be useful, just might lengthen the master cron job is all.

PR #184, opened 2018-08-17
Fixing xenial gazebo Dockerfiles

ahcorde:
I think we need to modify the base image of these Dockerfiles.

PR #185, opened 2018-08-17
Use os specific tags in gazebo base images

mikaelarguedas:
Now that some gazebo images target multiple Ubuntu distributions, we need to add the ability of depending on the os specific tag rather than the generic short tag in the generated Dockerfiles.

Should fix and superseed https://github.com/osrf/docker_images/pull/184


mikaelarguedas:
Upstream PR: https://github.com/docker-library/official-images/pull/4743

Issue #171, opened 2018-07-11
Why do the Dockerfiles stipulate specific version numbers?


lag-linaro:
At the moment, version numbers for varying architectures are not in sync.

Taking Gazebo8 and Xenial as an example, ARM64 and ARMHF are currently on 8.0.0 and AMD64 and I386 are on 8.6.0.  The Dockerfile for this combination stipulates 8.6.0, ensuring breakage when attempting to build images for architectures which should otherwise be supported.

mikaelarguedas:
thanks @lag-linaro for pointing this out.
The reason we pin the versions down is to ensure the images are rebuilt when a new version comes out (otherwise the cache is not busted and the images are not up to date).

The initial logic was developped for ROS packages where the versions are always in sync, and you are totally right, this doesnt apply to Gazebo and the logic should be changed to cover that case.

This seem like another good use case for either specifying exact version numbers on a per architecture basis (https://github.com/osrf/docker_images/issues/112) or maintaining separate dockefiles per architecture (https://github.com/osrf/docker_images/pull/115)

@ruffsl FYI

ruffsl:
As @mikaelarguedas duly pointed out, this ensures we always have a reliable mechanism for breaking the build cache when bumping up the version. It also affords us a degree of repeatability and regulation of the exact release that goes into the official images, and thus a cautionary procedure we abide by:  
https://github.com/docker-library/official-images#repeatability


Now that Docker Hub official libraries support multiarch:  
https://github.com/docker-library/official-images#multiple-architectures

Its doable to extend gazebo image builds for more than just amd64. When I first ported the ROS1 images to multiarch I briefly thought about doing the same for gazebo as well, but wasn't sure if there would be any user demand to justify the extension.

@lag-linaro , may I inquire you use case or applications that necessitating containerised gazebo on arm?

lag-linaro:
@mikaelarguedas, @ruffsl, thank you for taking the time to explain why you pin exact versions to builds in your Dockerfile and for being receptive to multi-arch builds.

@ruffsl in answer to your question, I'm afraid I do not have anything exciting to tell you.  As an integrator and ARM64v8 advocate/evangelist, I'm seeking to ensure that as many useful Docker Hub based projects are accessible to users of the latest hardware to be operated on the desktop and in the data centre.  Essentially I am just the middleman. :smile:

ruffsl:
Tracking arm support for gazebo here: https://github.com/osrf/docker_images/issues/173

PR #191, opened 2018-09-17
minor fixups to the docs

mikaelarguedas:
Updating the current parameters supported ans fix typos here and there

### DIFF
--- a/doc/maintaining_repo.md
+++ b/doc/maintaining_repo.md
@@ -6,22 +6,22 @@ The Dockerfiles used in Offical Docker Hub repos for OSRF projects are hosted he
  * Currently tags are nested to take advantage of the caching capabilities of docker. In this way, each tag build from each other like russan dolls. This is so that a user may still find a minimal tags that only include what they need, but should they eventually need more, no duplicate binaries need be downloaded or stored to disk.
 
 * Versions
- * The way project Debian packages are installed must include the specific package version number. This is to ensure that the version to be installed is explicitly stated and will not happen to change over the course of the domino effect of automated build stages. This also helps keep builds repeatable and reproducible, and is a good practice to fallow when making your own tags from the official images.
+ * The way project Debian packages are installed must include the specific package version number. This is to ensure that the version to be installed is explicitly stated and will not happen to change over the course of the domino effect of automated build stages. This also helps keep builds repeatable and reproducible, and is a good practice to follow when making your own tags from the official images.
 
 ## Update Procedure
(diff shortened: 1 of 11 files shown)


doc/maintaining_repo.md
```diff
@@ -6,22 +6,22 @@ The Dockerfiles used in Offical Docker Hub repos for OSRF projects are hosted he
...
 
-> This can be done simply however by using the scripts described in accompanying docs to regenerate all Dockerfiles. For ROS, the scripts will use ros_buildfarm.common package to fetch the latest package index and append the proper version numbers.
+> This can be done simply however by using the scripts described in accompanying docs to regenerate all Dockerfiles. For ROS, the scripts will use `ros_buildfarm.common` package to fetch the latest package index and append the proper version numbers.
 
-Also, we now pin the version used for each ROS package. Note that however, we are using reprepro (a tool to handle local repositories of debian packages), and as in FAQ 3.1 is limited to one version per architectur. So from the Repeatability documentation referenced
+Also, we now pin the version used for each ROS package. Note that however, we are using reprepro (a tool to handle local repositories of debian packages), and as in FAQ 3.1 is limited to one version per architecture.
```
mikaelarguedas: @ruffsl Not sure what the last sentence was trying to communicate (so I removed it). Feel free to push directly to this branch if you think a sentence should be added back

ruffsl: Looks like a trailing sentence I forgot about.

I probably meant to say something along the line of repeatability like with what I commented/linked in  https://github.com/osrf/docker_images/issues/171#issuecomment-404290452 .

gazebo/create_dockerfiles.py
```diff
@@ -24,19 +24,19 @@ def main(argv=sys.argv[1:]):
     parser.set()
     args = parser.parse(argv)
 
-    # Read platform perams
+    # Read platform params
```
mikaelarguedas: @ruffsl Maybe we should deduplicate these files as they seem to be identical between gazebo ros and ros2.
Do you expect these to diverge in the future or would it be safe to consolidate them ?

ruffsl: Originally they were each slightly different, but after refactoring the case specific logic into docker_templates lib, they should be the same. I think the CI makes the assumption that the scripts are executed in place, but we could probably just move `create_dockerfiles.py` `create_dockerfolders.py` and `create_dockerlibrary.py` to the root of the repo (e.g. `/scripts`) and just symlink them to each `ros`, `ros2`, `gazebo` folder.

Issue #201, opened 2018-10-19
Can we get a ros2:latest dockerfile?


mkhansen-intel:
We're trying to use Travis CI to test our Navigation code, and when building from source it takes too long (> 50 minutes) to build ROS2. We've been using the ros2:bouncy-desktop tagged docker image, but we'd like to move to using the master branch (or equivalent) for our development and testing. I wanted to know if there's a daily build (or at least weekly) that could generate a ros2:latest docker image for us to pull and use in our testing. That would really help.

sloretz:
I don't think there is a nightly docker image, but there are nightly packaging jobs. They create an archive with a ROS 2 workspace that could be used as an underlay for your CI

https://ci.ros2.org/view/packaging/

For a linux job
```bash
wget https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
tar -xjvf ros2-package-linux-x86_64.tar.bz2
. ros2-linux/setup.bash
# Now build other stuff on top
```

mkhansen-intel:
I'll give this a try! Thanks @sloretz

mkhansen-intel:
OK, this works. However, it still would be nicer if the nightly build also generated a Docker image like the ones being generated by this repo. Maybe a future enhancement? :) @ruffsl - is that something you would know how to do?

ruffsl:
@mkhansen-intel , to be clear, between the title of this issue vs what your requesting. are you asking for:
* A added ROS2 Dockerfile that builds the latest master branches of ROS2 from source
* Additionally, host a nightly derived Docker image using a automated repo on Docker Hub

Having an automated repo build the master branch of ROS2 from source wouldn't be hard, given we already have the Dockerfile scaffolding for creating one that builds the latest stable release from source:

https://github.com/osrf/docker_images/blob/55f03de3a06991c28a897982652dbafcb09d5592/ros2/source/source/Dockerfile#L66-L67

However, as per @sloretz mentioned, there seem to be some nice readily available nightly builds from https://ci.ros2.org/view/packaging, so I would be inclined to rely on those, rather than force Travis or Docker Hup to duplicate the same workload.

[...]

PR #202, opened 2018-10-20
WIP | Adding ros2 nightly dockerfile

ruffsl:
This PR address https://github.com/osrf/docker_images/issues/201 and addes a new Docker Image tag, 'nightly', to the ROS2 automated repo, providing a covenant installation of ci.ros2.org nightly builds. Some remaining action items:

- [ ] Create respective docker_template file for nightly dockerfile
- [ ] Parameterized dockerfile elements into images.yaml.em
- [ ] Reduce image size (prune unneeded apt install?)
- [ ] Add webhook for new tag on Docker Hub
- [ ] Trigger webhook from ci.ros2.org nightly job

ping @mkhansen-intel

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -1,0 +1,85 @@
+# This is an auto generated Dockerfile for ros2:nightly
+# generated from docker_images/create_ros2_nightly_image.Dockerfile.em
+FROM ubuntu:bionic
+
+# set timezone
+RUN echo 'Etc/UTC' > /etc/timezone && \
+    ln -s /usr/share/zoneinfo/Etc/UTC /etc/localtime && \
+    apt-get update && apt-get install -q -y tzdata && rm -rf /var/lib/apt/lists/*
+
+# install packages
+RUN apt-get update && apt-get install -q -y \
+    bash-completion \
+    dirmngr \
+    git \
+    gnupg2 \
+    libasio-dev \
+    libtinyxml2-dev \
+    lsb-release \
+    python3-pip \
+    wget \
+    && rm -rf /var/lib/apt/lists/*
+
+# setup keys
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 421C365BD9FF1F717815A3895523BAEEB01FA116
+
+# setup sources.list
+RUN . /etc/os-release \
+    && echo "deb http://repo.ros2.org/$ID/main `lsb_release -sc` main" > /etc/apt/sources.list.d/ros2-latest.list
+
+# setup environment
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
+
+# install packages from the ROS repositories
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python3-catkin-pkg-modules \
+    python3-colcon-common-extensions \
+    python3-rosdep \
+    python3-vcstool \
+    && rm -rf /var/lib/apt/lists/*
+
+# install python packages
+RUN pip3 install -U \
+    argcomplete \
+    flake8 \
+    flake8-blind-except \
+    flake8-builtins \
+    flake8-class-newline \
+    flake8-comprehensions \
+    flake8-deprecated \
+    flake8-docstrings \
+    flake8-import-order \
+    flake8-quotes \
+    pytest-repeat \
+    pytest-rerunfailures
+
+# bootstrap rosdep
+ENV ROSDISTRO_INDEX_URL https://raw.githubusercontent.com/ros2/rosdistro/ros2/index.yaml
+RUN rosdep init \
+    && rosdep update
+
+# download nightly build
+ENV ROS2_INSTALL /opt/ros/nightly
+RUN mkdir -p $ROS2_INSTALL
+WORKDIR $ROS2_INSTALL
+
+ENV ROS2_BINARY_URL https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
+RUN wget -q $ROS2_BINARY_URL -O - | \
+    tar -xj --strip-components=1 \
+      ros2-linux -C $ROS2_INSTALL
+
+# install dependencies
+RUN apt-get update && rosdep install -y \
+    --from-paths $ROS2_INSTALL/share \
+    --ignore-src \
+    --rosdistro bouncy \
+    --skip-keys "console_bridge fastcdr fastrtps libopensplice67 rti-connext-dds-5.3.1 urdfdom_headers" \
+    && rm -rf /var/lib/apt/lists/*
+
+# setup entrypoint
+COPY ./ros2_nightly_entrypoint.sh /ros2_entrypoint.sh
+
+WORKDIR /root
+ENTRYPOINT ["/ros2_entrypoint.sh"]
+CMD ["bash"]
--- a/ros2/nightly/Makefile
+++ b/ros2/nightly/Makefile
@@ -1,0 +1,19 @@
+all: help
+
+help:
+	@echo ""
+	@echo "-- Help Menu"
+	@echo ""
+	@echo "   1. make build            - build all images"
+	@echo "   2. make pull             - pull all images"
+	@echo "   3. make clean            - remove all images"
+	@echo ""
+
+build:
+	@docker build --tag=ros2:nightly			nightly/.
+
+pull:
+	@docker pull ros2:nightly
+
+clean:
+	@docker rmi -f ros2:nightly
--- a/ros2/nightly/images.yaml.em
+++ b/ros2/nightly/images.yaml.em
@@ -1,0 +1,50 @@
+%YAML 1.1
+# ROS2 Dockerfile database
+---
+images:
+    nightly:
+        base_image: @(os_name):@(os_code_name)
+        maintainer_name: @(maintainer_name)
+        template_name: docker_images/create_ros2_nightly_image.Dockerfile.em
+        entrypoint_name: docker_images/ros2_nightly_entrypoint.sh
+        template_packages:
+            - docker_templates
+        upstream_packages:
+            - bash-completion
+            - git
+            - libasio-dev
+            - libtinyxml2-dev
+            - wget
+        ros2_repo_packages:
+            - python3-catkin-pkg-modules
+            - python3-rosdep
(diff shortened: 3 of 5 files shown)


ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+    flake8-comprehensions \
+    flake8-deprecated \
+    flake8-docstrings \
+    flake8-import-order \
+    flake8-quotes \
+    pytest-repeat \
+    pytest-rerunfailures
+
+# bootstrap rosdep
+ENV ROSDISTRO_INDEX_URL https://raw.githubusercontent.com/ros2/rosdistro/ros2/index.yaml
```
ruffsl: Is this still necessary?
I don't see this in the wiki anymore:
https://github.com/ros2/ros2/wiki/Linux-Install-Binary#installing-and-initializing-rosdep

ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+RUN wget -q $ROS2_BINARY_URL -O - | \
+    tar -xj --strip-components=1 \
+      ros2-linux -C $ROS2_INSTALL
+
+# install dependencies
+RUN apt-get update && rosdep install -y \
+    --from-paths $ROS2_INSTALL/share \
+    --ignore-src \
+    --rosdistro bouncy \
+    --skip-keys "console_bridge fastcdr fastrtps libopensplice67 rti-connext-dds-5.3.1 urdfdom_headers" \
```
ruffsl: This list of skipped keys is a bit shorter than the wiki,
should we prune this a little, or would it be helpful to keep.
https://github.com/ros2/ros2/wiki/Linux-Install-Binary#installing-the-missing-dependencies

ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+    pytest-repeat \
+    pytest-rerunfailures
+
+# bootstrap rosdep
+ENV ROSDISTRO_INDEX_URL https://raw.githubusercontent.com/ros2/rosdistro/ros2/index.yaml
+RUN rosdep init \
+    && rosdep update
+
+# download nightly build
+ENV ROS2_INSTALL /opt/ros/nightly
```
ruffsl: We could have the template have this chance the latest distro name, but I think keeping it static to nightly might be helpful.

ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+RUN mkdir -p $ROS2_INSTALL
+WORKDIR $ROS2_INSTALL
+
+ENV ROS2_BINARY_URL https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
```
ruffsl: This could become a snippet in the template

ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+# install dependencies
+RUN apt-get update && rosdep install -y \
+    --from-paths $ROS2_INSTALL/share \
+    --ignore-src \
+    --rosdistro bouncy \
+    --skip-keys "console_bridge fastcdr fastrtps libopensplice67 rti-connext-dds-5.3.1 urdfdom_headers" \
+    && rm -rf /var/lib/apt/lists/*
+
+# setup entrypoint
+COPY ./ros2_nightly_entrypoint.sh /ros2_entrypoint.sh
```
ruffsl: I think renaming the entrypoint to be consistent with the stable releases should help portability, e.g. when switching between release and nightly images.

ros2/nightly/nightly/Dockerfile
ruffsl: Perhaps we could set `ROS_DISTRO` to `nightly`, and use the traditional entrypoint; but perhaps that might throw off other ros tooling.

ruffsl:
@nuclearsandwich or @tfoote , would you know how to integrate that last step:
> Trigger webhook from ci.ros2.org nightly job

I can show you the token/url at dockerhup that a cron job would have to poke to trigger the automated docker build.

ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+    && rosdep update
+
+# install python packages
+RUN pip3 install -U \
+    argcomplete
+
+# install ros2 packages
+ENV ROS_DISTRO crystal
+RUN mkdir -p /opt/ros/$ROS_DISTRO
+ENV ROS2_BINARY_URL https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
```
nuclearsandwich: Is this environment variable useful at runtime or is it better to make it a build arg
```suggestion
ARG ROS2_BINARY_URL https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
```

ros2/nightly/nightly/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+RUN wget -q $ROS2_BINARY_URL -O - | \
+    tar -xj --strip-components=1 -C /opt/ros/$ROS_DISTRO
+RUN sed -i "s|^\(_colcon_prefix_chain_sh_COLCON_CURRENT_PREFIX\s*=\s*\).*$|\1/opt/ros/$ROS_DISTRO|" \
+      /opt/ros/$ROS_DISTRO/setup.sh && \
+    sed -i "s|^\(_colcon_prefix_sh_COLCON_CURRENT_PREFIX\s*=\s*\).*$|\1/opt/ros/$ROS_DISTRO|" \
+      /opt/ros/$ROS_DISTRO/local_setup.sh
+
+# ignore specified rosdep keys
+COPY core.yaml /etc/ros/rosdep/
```
nuclearsandwich: Is there a reason to inject these empty rosdeps rather than running with `--skip-keys` as the [instructions](https://index.ros.org/doc/ros2/Installation/Linux-Install-Binary/#id5) do? Since the list is maintained by hand a list of skipped keys would seem to me to be only as much work to maintain as the core.yaml.

nuclearsandwich: Answering my own question, the reason is so that "downstream" invocations of rosdep benefit from treating the same packages as resolved rather than having to propagate the skip list. :cool:.

ros2/nightly/nightly/core.yaml
```diff
@@ -0,0 +1,15 @@
+console_bridge: {ubuntu: []}
+fastcdr: {ubuntu: []}
+fastrtps: {ubuntu: []}
+libopensplice67: {ubuntu: []}
+libopensplice69: {ubuntu: []}
```
nuclearsandwich: Using an empty list to satisfy these will require anyone who does want to test with OpenSplice (or RTI Connext DDS below) to explicitly install the package as rosdep will resolve to the empty package.

nuclearsandwich:
> I can show you the token/url at dockerhup that a cron job would have to poke to trigger the automated docker build.

@ruffsl and I spoke offline about this. As an initial rollout I'll have a cron-job that's set to run two hours after the ci build is triggered which checks the status of the latest build and triggers a docker build on success. If that works well we'll move that cron job's home to ROS 2 CI. This is mostly to avoid managing the Docker Hub token securely and handling the fact that only ci.ros2.org and not a community deployment of our CI scripts ought to be triggering builds.

ros2/nightly/nightly/core.yaml
ruffsl: Do you know of a alternate way with could avoid installing these transport venders without this injection? Is this list overcompensating, or is there only a smaller set of keys we just need to skip?

> to explicitly install the package as rosdep will resolve to the empty package.

Users could also add a new files to `/etc/ros/rosdep/sources.list.d/` that takes president, no? Something like `10-default.list`, denoting priority like with udev rules? I thought of adding the the rosdep entries for the `prereqs.yaml` and `ros2.yaml` into a separate `default.list` file with higher priority, but then decided against deviating from the existing footprint of a conventional install for the sake of consistency.

nuclearsandwich: > Is this list overcompensating, or is there only a smaller set of keys we just need to skip?

The list in the documentation is a bit overstuffed to work for both Bouncy and Crystal. In general rosdep doesn't have the ability to distinguish between an optional or opportunistic dependency (use this this if it's there but don't freak out if it isn't) and a mandatory dependency. So I don't know that there's anything to do about it. It's just something that downstream consumers of the Dockerfile need to know if their intent is to use OpenSplice or RTI Connext DDS.

ros2/nightly/nightly/Dockerfile
nuclearsandwich: > I don't suspect this list changes frequently?

It changes when we introduce new prerequisites, which yeah, isn't often.

ros2/nightly/nightly/core.yaml
ruffsl: Suppose we use `--skip-keys` for the opportunistic dependencies and keep only the mandatory dependencies in the prereqs.yaml, would that satisfy both uses? Short of calling rosdep install on a core package, rosdep wouldn't necessarily try to install the opportunistic dependency from a parent package already ignored/installed on the system.

ruffsl:
> As an initial rollout I'll have a cron-job that's set to run two hours after the ci build is triggered which checks the status of the latest build and triggers a docker build on success.

I checked the build history of the docker hub repo, and haven't seen any triggers for the last 5 days. I tested the ruby script you shared offline on my host, and it triggered. Are you sure your cron job up and scheduled?

ros2/nightly/nightly/core.yaml
ruffsl: See commit ea0fdab

ruffsl:
Ping @nuclearsandwich , I still don't see the automate DockerHub repo being triggered by that cron-job. I've just been waiting to merge this until that last bit of the pipeline is working.

nuclearsandwich:
> I still don't see the automate DockerHub repo being triggered by that cron-job. I've just been waiting to merge this until that last bit of the pipeline is working.

Hah. I didn't think it would work until after the PR merged so I've been waiting to turn it on :sweat_smile:. I'll start it tonight and hopefully we'll see it tomorrow.

Issue #197, opened 2018-10-08
Preserve snapshot version of images?


130s:
Has there been any demand/plan to preserve "snapshot" images, e.g. images taken at the certain period that will basically never be updated?

This would allow us to take a control over the version of all packages on the container.

I believe those who build the products upon Docker images somehow maintain the version of the packages inside the image and/or the base image version.

On [Ubuntu's hub](https://hub.docker.com/r/library/ubuntu/tags) there are such images. E.g. [xenial-20180808](https://hub.docker.com/r/library/ubuntu/tags/xenial-20180808/).

(CC @AustinDeric, [ros-industrial/docker](https://github.com/ros-industrial/docker) repo maintainer.)

sloretz:
I'm not sure if there has been discussion about adding tags with a date in the name. I would recommend tagging the image that should remain frozen instead.

```
sudo docker pull osrf/ros2:bouncy-core
sudo docker tag osrf/ros2:bouncy-core yourdockerhubusername/ros2:bouncy-core-20181008
sudo docker push yourdockerhubusername/ros2:bouncy-core-20181008
```

ruffsl:
@130s , I suspect you would like to pin by its digest, or immutable identifier:

### [Pull an image by digest (immutable identifier)](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier)

> In some cases you don’t want images to be updated to newer versions, but prefer to use a fixed version of an image. Docker enables you to pull an image by its digest. When pulling an image by digest, you specify exactly which version of an image to pull. Doing so, allows you to “pin” an image to that version, and guarantee that the image you’re using is always the same.

See this related comment here for more details: https://github.com/osrf/docker_images/issues/104#issuecomment-343583136

If you to dig through the archives to trace back any particular build (or "snapshot"), you may want to check out the Extended Information repo for the docker library: [docker-library/repo-info](https://github.com/docker-library/repo-info).

Specifically, for ros:

https://github.com/docker-library/repo-info/tree/master/repos/ros

> Maybe taking a snapshot per every public sync of each distro and push the image to the public registry.

In theory, I think the `remote` folder in the repo above should only receive new commits when there automated CI generates a new report after a tagged image in the library is updated. Aside from triggers from upstream images of debian or ubuntu, new images should only be pushed to the library when we update the versions of the metapackages, that our own CI is monitoring for. New releases traditionally bump these meta package versions, no?

[...]

Issue #125, opened 2018-02-16
Enable PR bot to create upstream manifest PRs from local merge events


ruffsl:
Context in https://github.com/docker-library/official-images/issues/3890#issuecomment-357102830

It would be nice to automate the chained process of triggering an upstream PR to docker-library/official-images to update those manifests upon meaning one of @osrf-docker-builder PR locally. Given that @osrf-docker-builder is controlled through the travis CI, I'm unaware of the best method to capture github merge events for triggering further github API calls for PR creation. Are there proper methods to reactively do this, or should we just rely on the merge commits to master triggering a new CI job? I'd still like to control when upstream PRs are created though, say by keywords or dialog in PR merge message.

PR #204, opened 2018-11-10
Update Dockerfile with template for breaking build cache

ruffsl:
Fixes: https://github.com/osrf/docker_images/issues/112

### DIFF
--- a/gazebo/7/ubuntu/xenial/gzserver7/Dockerfile
+++ b/gazebo/7/ubuntu/xenial/gzserver7/Dockerfile
@@ -17,8 +17,13 @@ RUN . /etc/os-release \
     && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list
 
 # break build cache for sync
-RUN echo "Release: Fri, 09 Nov 2018 14:51:50 UTC" \
-    && echo "Digest: 1b1a9870d531fcfee20ae1ab3df60d499d6b206c11fe30e3c89dd111ae31c756"
+RUN apt-get update \
+    && . /etc/os-release \
+    && echo "Release: Fri, 16 Nov 2018 16:26:55 UTC" \
+    && export SUM=021bafbd47748fa31505efcf66f499f8f9d7601c18f6d8456ec76686b76711df \
+    && export FILE=/var/lib/apt/lists/packages.osrfoundation.org_gazebo_$ID-stable_dists_$(lsb_release -sc)_InRelease \
+    && echo "$SUM *$FILE" | sha256sum --check \
+    && rm -rf /var/lib/apt/lists/*
 
 # install gazebo packages
 RUN apt-get update && apt-get install -q -y \
--- a/gazebo/7/ubuntu/xenial/gzserver7/Dockerfile
+++ b/gazebo/7/ubuntu/xenial/gzserver7/Dockerfile
@@ -16,6 +16,10 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD8
 RUN . /etc/os-release \
     && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list
 
+# break build cache for sync
+RUN echo "Release: Fri, 09 Nov 2018 14:51:50 UTC" \
+    && echo "Digest: 1b1a9870d531fcfee20ae1ab3df60d499d6b206c11fe30e3c89dd111ae31c756"
+
 # install gazebo packages
 RUN apt-get update && apt-get install -q -y \
     gazebo7=7.14.0-1* \
--- a/gazebo/8/ubuntu/xenial/gzserver8/Dockerfile
+++ b/gazebo/8/ubuntu/xenial/gzserver8/Dockerfile
@@ -17,8 +17,13 @@ RUN . /etc/os-release \
     && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list
 
 # break build cache for sync
-RUN echo "Release: Fri, 09 Nov 2018 14:51:50 UTC" \
-    && echo "Digest: 1b1a9870d531fcfee20ae1ab3df60d499d6b206c11fe30e3c89dd111ae31c756"
+RUN apt-get update \
+    && . /etc/os-release \
+    && echo "Release: Fri, 16 Nov 2018 16:26:55 UTC" \
+    && export SUM=021bafbd47748fa31505efcf66f499f8f9d7601c18f6d8456ec76686b76711df \
+    && export FILE=/var/lib/apt/lists/packages.osrfoundation.org_gazebo_$ID-stable_dists_$(lsb_release -sc)_InRelease \
+    && echo "$SUM *$FILE" | sha256sum --check \
+    && rm -rf /var/lib/apt/lists/*
 
 # install gazebo packages
 RUN apt-get update && apt-get install -q -y \
--- a/gazebo/8/ubuntu/xenial/gzserver8/Dockerfile
+++ b/gazebo/8/ubuntu/xenial/gzserver8/Dockerfile
@@ -16,6 +16,10 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD8
 RUN . /etc/os-release \
     && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list
 
+# break build cache for sync
+RUN echo "Release: Fri, 09 Nov 2018 14:51:50 UTC" \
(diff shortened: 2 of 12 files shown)


ruffsl:
@tianon or @yosifkit , is this approach reasonable?
Context: https://github.com/osrf/docker_images/issues/112#issuecomment-437556475

gazebo/7/ubuntu/xenial/gzserver7/Dockerfile
```diff
@@ -16,6 +16,10 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD8
 RUN . /etc/os-release \
     && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list
 
+# break build cache for sync
+RUN echo "Release: Fri, 09 Nov 2018 14:51:50 UTC" \
+    && echo "Digest: 1b1a9870d531fcfee20ae1ab3df60d499d6b206c11fe30e3c89dd111ae31c756"
```
tianon: Instead of just a useless `echo` to do a cache bust, here's what I had in mind:
```dockerfile
# This is an auto generated Dockerfile for gazebo:gzserver7
# generated from docker_images/create_gzserver_image.Dockerfile.em
FROM ubuntu:xenial

# install packages
RUN apt-get update && apt-get install -q -y \
    dirmngr \
    gnupg2 \
    lsb-release \

# setup keys
RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD83DB69272AFE98867170598AF249743

# setup sources.list
RUN . /etc/os-release \
    && echo "deb http://packages.osrfoundation.org/gazebo/$ID-stable `lsb_release -sc` main" > /etc/apt/sources.list.d/gazebo-latest.list

# install gazebo packages
RUN apt-get update \
    && . /etc/os-release \
    && echo "021bafbd47748fa31505efcf66f499f8f9d7601c18f6d8456ec76686b76711df */var/lib/apt/lists/packages.osrfoundation.org_gazebo_$ID-stable_dists_$(lsb_release -sc)_InRelease" | sha256sum -c - \
    && apt-get install -q -y \
        gazebo7=7.14.0-1* \

# setup environment
EXPOSE 11345

# setup entrypoint
COPY ./gzserver_entrypoint.sh /

ENTRYPOINT ["/gzserver_entrypoint.sh"]
CMD ["gzserver"]
```

(Actually verifying the hash of the interesting file in the appropriate place.)

ruffsl: Ok @tianon , I've adjusted the break point to check the digest of the InRelease file.
I didn't insert it inline with the last install step, as I wanted to the break point to be early, just after the repo is added, so that the rest of repo packages have a better chance staying in sync.

130s:
Is there anywhere I can look at the result of this change?
I'm not super knowledgeable about this sort of stuff so can't give code-review, but still willing to contribute regarding the result of this change (I came from https://github.com/osrf/docker_images/issues/197).

ruffsl:
> Is there anywhere I can look at the result of this change?

@130s , could you clarify what you mean by "result"?

Do you me to ask:
* the change to the resulting images produced?
* the change to the status quo of the pipeline of PR wrt this repo?

The docker images produced would resulting include a new layer that would have the history of this InRelease check. So the metadata about the release this was built with could be inferred. It wouldn't be as obvious as reading an environment variable or checking the image tag. I briefly considered have the template set an `ENV` of the time and/or digest to break the catch, but I'm hesitant to pollute downstream user environment (@tianon if there is a convention or precedent for this already in the library, I'd take a look).

The other thing that 8e4a5376f204a270edb4e95a5fa4464c8954b3c9 does in addition to 03b69b706454258afc39a45e5344ad3aaf6a71a8 is that it will actually assert that the InRlease in the image matches the the digest templated into the dockerfile, and return non zero otherwise. So in addition to the version of the relevelt pined package bumping, a sync to any package in the package repo will break the build, helping to more tightly preserve the [repeatability](https://github.com/docker-library/official-images#repeatability) of the Dockerfile.

This will effectively block newer builds of an image until the Dockerfile itself is updated to reflect the latest sync until the CI changes are propagated upstream. Still, I'm not sure this is best practice, as the blocking might slow down the propagation of upstream security patches inherited from base images (e.g. ubuntu, debian, etc). My ideal level of repeatability would be a little loser in that pinged packaged would be consistent across consecutive builds of a dockerfile, but the unpinned supported packages could be subject to upstream security updates.

This is why I first opted for just a passive echo in 03b69b706454258afc39a45e5344ad3aaf6a71a8 , just to make a change in the dockerfile to break the build cache, but not to necessary block future builds. I'd say until https://github.com/osrf/docker_images/issues/125 is resolved, and we have a more streamlined end-to-end PR pipeline, I'd be hesitant in merging 03b69b706454258afc39a45e5344ad3aaf6a71a8, as the demand on maintainers here would increase given we'd more frequently break the critical build path of security updates with any repo sync, regardless if relevant pinned packages where actually bumped. 

@nuclearsandwich  is there an easy way now of having bots trigger new PRs by being triggered from other PR's being merged. https://github.com/osrf/docker_images/issues/125 would benefit from having something ensure the upstream manifest files always stays in sync with copy of the manifest we generate locally in this repo.

nuclearsandwich:
> is there an easy way now of having bots trigger new PRs by being triggered from other PR's being merged.

Can you draw me a little diagram of what you envision here. What upstream repos are you trying to track changes in?

ruffsl:
> This is pretty frequent

Indeed, that's why I'd like further automate the PR process upstream. I think we could use CircleCI's Manual Job Approval in a Workflow to allows us to control and deployed upstream PRs, so that our bots don't submit upstream PRs unsupervised, while still not burdening approvers with having to open them ourselves. Then we would be able to keep with the the relatively frequent syncs.

https://circleci.com/blog/manual-job-approval-and-scheduled-workflow-runs/

@sloretz , I can add the circleci logic, but would you be able to add another python script like we have currently controlling our PR bot to make PRs targeting upstream master based from our manifest in local master?

sloretz:
> I can add the circleci logic, but would you be able to add another python script like we have currently controlling our PR bot

I can do that. Where's the script powering the PR bot?

> to make PRs targeting upstream master based from our manifest in local master?

As in, a script that creates a PR like [this one](https://github.com/docker-library/official-images/pull/5280/files)? Is this script also updating the hashes and creating the dockerfiles? How should it get the github token from the context it's run in? Keyring? Env var?

ruffsl:
> I can do that. Where's the script powering the PR bot?

The current code for the PR bot resides in the travis.py file here:

https://github.com/osrf/docker_images/blob/99b90d8b37833010eb2aa72a517ba4c35c0830e8/.travis/travis.py#L241-L246

> As in, a script that creates a PR like [this one](https://github.com/docker-library/official-images/pull/5280/files)?

Yep, it'll need to generalized for both the `ros` and `gazebo` manifests.
The travis.py script does this by checking the HUB_REPO env.

> Is this script also updating the hashes and creating the dockerfiles?

The script/feature needed doesn't have to update the hashes in the local manifests, this is already achieved by the create_docker_library script that are created by the existing PR CI. What this script should do is check if the respective manifest differs from that of upstream and open a github PR with some kind of bot template if they are out of sync. 

> How should it get the github token from the context it's run in? Keyring? Env var?

You can check the from env vars like the existing travis.py one does, and we'll just configure them into the CI admin panel as private variables.

The idea in general: I'm thinking we could have the existing CI as-is, that just continues to open PR for when dockerfiles and manfest are generated out of date. The added circleci would additionally trigger on each PR and the workflow would immediately block waiting on maintainer confirmation. The maintainer checks that the existing CI can build and pass the generated PR, then clicks the github merge button, then click the CircleCI link to approve button to deploy the upstream PR, where the circleci workflow continues and runs the new script that diffs the respective manifest the local PR just merged wrt upstream. If they differ, it then check if an existing template PR is already open, if so it could error out, if not it opens a new one up using a template that @pings the maintainer group.

mikaelarguedas:
I believe this was discussed in the tickets, but I'd like to highlight that this means **many more** invalidations and rebuilds than the other suggested approaches, situations that come to mind:
- the release of a python package will invalidate **all** ROS (1 and 2) images (e.g. `python3-colcon-*` that get new releases pretty often)
- a sync of a specific distro will invalidate images of other distributions, e.g a ROS Lunar sync will result in the ROS Kinetic xenial and ROS Melodic stretch invalidation
- the sync of a ROS 1 distro will invalidate images of ROS 2 distros: e.g. a ROS Melodic sync will invalidate the Bouncy, Crystal and Dashing images (and likely ROS2 E)
- release of any leaf package not part of desktop_full will invalidate every image of that ROS distro and other ROS distro targetting the same OS.

As ROS2 now also installs always a single package, it may be not too cluttering to just provide the version number suggested in https://github.com/osrf/docker_images/issues/112#issuecomment-361378401 and avoid invalidating when not necessary.

---

I agree that automated builds and PR creation would be very beneficial if this is the strategy adopted :+1:.
It may be worth considering if there are ways to reduce the number of automated PRs (both here and upstream) as well as manual edits needed when new distros (ROS or Ubuntu/Debian) come out.

sloretz:
@ruffsl Added script to create upstream PRs if the librarry definition files are different in 8931512f43d8174a8fed58cc929297ec654d96d0

I pushed it to the branch in #230

mikaelarguedas:
Superseded by https://github.com/osrf/docker_templates/pull/90, example or resulting automatic PR: https://github.com/osrf/docker_images/pull/459

PR #224, opened 2019-02-13
update index.yaml to use ros1 rosdistro

Karsten1987:
the original link is not working any longer, results in 404 and thus the dockerfile doesn't complete.


sloretz:
Ooops, I merged this too quick. The line that should be changed is in [this template](https://github.com/osrf/docker_images/blob/443867a27622f82669abcfab7d1b24bc896182c9/ros2/source/images.yaml.em#L42). I'll make a follow up pr

PR #226, opened 2019-02-17
[ros2/source] Skip libopensplice69 key

mikaelarguedas:
Otherwise opensplice is installed and message libraries are built for it


tfoote:
I think older images are still using it.

mikaelarguedas:
Yeah I kept it to match the [installation instructions](https://index.ros.org/doc/ros2/Installation/Linux-Development-Setup/#id7) but it's not needed for crystal, only older distros. The current list should allow to build crystal or bouncy by just changing the repos file URL

PR #227, opened 2019-02-18
EOL ardent

mikaelarguedas:
EOL ardent docker images as the distribution is EOL since December and the buildfarm doesn't build packages for it anymore


mikaelarguedas:
PR to official library: https://github.com/docker-library/official-images/pull/5460

PR #230, opened 2019-02-28
Release hash script and update hashes

sloretz:
@ruffsl starting from `release_hash` in #204 this merges master into itself, adds a python script to update the `InRelease` file hashes, and updates the hashes. If it looks ok to you feel free to merge this into `release_hash`.

The script does not change the date in the echo'd string. Maybe it could echo the hash instead?


update_cache_busting_hashes.py
```diff
@@ -0,0 +1,80 @@
...
+    ]
+
+    url_hashes = {}
+    for url in urls:
+        hash_output = get_sha256_sum(url)
+        url_hashes[url] = hash_output
+        print(url)
+        print("\t" + hash_output)
+
+    paths = {
```
ruffsl: This enumeration of paths doesn't seem very maintaintable. I'd recommend we keep with template library updating the files, and simply checking if the diff of the file has changed compared to the commit history.

https://github.com/osrf/docker_templates/pull/45

pr_docker_library.py
```diff
@@ -0,0 +1,147 @@
...
+
+    fork_branch = update_library_definition_fork(name)
+
+    # Create pull request
+    create_upstream_pull_request(name, diff, fork_branch)
+
+
+if __name__ == '__main__':
+
+    update_upstream_library_definition('ros')
```
ruffsl: I think the switch for checking ros or gazebo should be inferred from `HUB_REPO` in CI environment , and should only seek to update the upstream manifest for which the automated PR pertains to.

https://github.com/osrf/docker_images/blob/cb0ce5bef267d8c305f6cfd2ded846c9444c4fc7/.travis/travis.py#L74

pr_docker_library.py
```diff
@@ -0,0 +1,147 @@
...
+    differ = difflib.Differ()
+    diff = differ.compare(
+        upstream_ldf.splitlines(keepends=True),
+        local_ldf.splitlines(keepends=True))
+    diff = ''.join(diff)
+    return diff
+
+
+def update_library_definition_fork(name):
+    # Create branch on fork that is up to date with upstream
```
ruffsl: Is this pushing the manifest (post merged) copied from osrf/docker_images to the `FORK_LIBRARY_REPO_URL` or only that from the checked out PR? To consolidate all the release changes for a given docker repo to one at a time, I think pushing the merged state of the repo manifest would be less overwhelming upstream that PRing each change. E.g. we'd only approve this automated upstream PR for the last merged PR to a given docker repo to batch release changes. This is similar to what we do now manually.

sloretz: I don't understand the question. Here is an example PR created by this script: ros-infrastructure/official-images#8 . Does the diff answer it?

update_cache_busting_hashes.py
sloretz: Ah, I missed that PR. I should have figured those sections were already created by a script.

sloretz: >  I'd recommend we keep with template library updating the files, and simply checking if the diff of the file has changed compared to the commit history.

I'm on board with `osrf/docker_templates` updating the files, but I don't understand what the second half of the sentence means. What checks the diff of what file compared to what commit history?

pr_docker_library.py
sloretz: 7fcdc9913cc1f73e263b06d1f7475b4045b9c1c9

sloretz: Will CI run this as a script or import the python function?

update_cache_busting_hashes.py
sloretz: Regardless, deleted this file in 815e3ff4133e9042324f2f71c01adb7b239fdeff

pr_docker_library.py
ruffsl: The CI will run this script, but only after the maintainer has manually clicked approve on CircleCI PR job for it to proceed.

pr_docker_library.py
ruffsl: Your pr is a multiple merged distro changes, yet local PRs by the @osrf-docker-builder separate each distro PR, so the maintainer should be prudent in waiting for all the changes targeting a given repo manifest to be merged before upstreaming the final updated manifest. To avoid opening to many PR, they are batched into one periodic PR for that respective dockerhub repo.

Example PR:
https://github.com/osrf/docker_images/pull/233
https://github.com/osrf/docker_images/pull/234
https://github.com/osrf/docker_images/pull/235
https://github.com/osrf/docker_images/pull/236
https://github.com/osrf/docker_images/pull/231

Upstream PR:
https://github.com/docker-library/official-images/pull/5573

update_cache_busting_hashes.py
ruffsl: > What checks the diff of what file compared to what commit history?

The script (that runs after the maintainer merges the last relevant PR and the approved this CI job) should first check if the PR itself results in changes to the manifest for the respective dockerhub repo. If PR changes the manifest file, it should then check if the manifest in `osrf/docker_images/blob/master/<repo>/<repo>` is out of data w.r.t `docker-library/official-images/blob/master/library/<repo>`. If the upstream manifest is out of sync, in should then check if no existing PR is open for the respective repo (perhaps checking the <repo> branch on our ros-infrastructure fork). If no existing PR exists, then it should push to the <repo> branch on our fork and open a templated PR, @ pinging us as maintainers.

mikaelarguedas:
I believe this can be closed in favor of https://github.com/osrf/docker_templates/pull/90

@ruffsl If there anything from the automatic PR work in this PR we want to keep in a new PR ?

ruffsl:
@mikaelarguedas , Looks like the librarians aren't a fan of  our sha lable tricks. Should we revert that and go back to this instead? The docker librarians seemed to like this approach as it would block un-reproducible builds by acting upon the discrepancies.
https://github.com/osrf/docker_images/pull/204#pullrequestreview-177031994

PR #242, opened 2019-03-20
Update rosdistro to track latest released distro

ruffsl:


ros2/source/images.yaml.em
```diff
@@ -42,7 +42,7 @@ images:
             install:
                 - --from-paths src
                 - --ignore-src
-                - --rosdistro bouncy
+                - --rosdistro crystal
```
mikaelarguedas: could we make the distro name a parameter that we will use on this line as well as for constructing the URL [L49](https://github.com/osrf/docker_images/blob/6a7c64ed48d742c41b5db64c168fa5e6e5ce6d31/ros2/source/images.yaml.em#L49)?

ruffsl: Not sure what you mean. Set `ENV ROS_DISTRO` and use that to `${}` substitute in the `githubusercontent.com/ros2/ros2/` URL? I thought we'd just pass the URL from the config outright so that users may point to custom gists and the like.

mikaelarguedas: It's a nitpick, I'm fine merging as is, I'm mostly trying to look for ways to avoid mismatches in the future

---

This PR was necessary because there was a mismatch between the distro used by rosdep and the repos file pulled.

If we used a url with a format like `https://raw.githubusercontent.com/ros2/ros2/@ros2distro/ros2.repos` and rosdep with something like:
```
            install:
                - --from-paths src
                - --ignore-src
                - --rosdistro @ros2distro
```
We would be ensured that the invocations match in the future.

ruffsl: I think that is good idea. Please see b6d5cde and a8955e9 .
I could update the template accordingly.

mikaelarguedas: looks good to me :+1: thanks for iterating

mikaelarguedas:
with https://github.com/osrf/docker_templates/pull/53 and https://github.com/osrf/docker_templates/pull/56 this should be good to go

PR #244, opened 2019-03-20
use templated ROS bootstrap tools install

mikaelarguedas:
This update the Dockerfiles based on https://github.com/osrf/docker_templates/pull/52

### DIFF
--- a/ros/bouncy/ubuntu/bionic/platform.yaml
+++ b/ros/bouncy/ubuntu/bionic/platform.yaml
@@ -10,4 +10,5 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version: 2
+    version:
+    ros_version: 2
--- a/ros/bouncy/ubuntu/bionic/platform.yaml
+++ b/ros/bouncy/ubuntu/bionic/platform.yaml
@@ -10,4 +10,4 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version:
+    version: 2
--- a/ros2/nightly/platform.yaml
+++ b/ros2/nightly/platform.yaml
@@ -10,4 +10,5 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version: 2
+    version:
+    ros_version: 2
--- a/ros2/nightly/platform.yaml
+++ b/ros2/nightly/platform.yaml
@@ -10,4 +10,4 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version:
+    version: 2
--- a/ros/.config/ros1/platform.yaml.em
+++ b/ros/.config/ros1/platform.yaml.em
@@ -9,5 +9,6 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version: 1
+    version:
     release: ros
+    ros_version: 1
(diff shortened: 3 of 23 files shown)


mikaelarguedas:
@ruffsl  I pushed the version with the updated dockerfile to reflect https://github.com/osrf/docker_templates/pull/52/commits/732a26914cc2cccb4879548abf54e0ebf658d136.
Would you mind giving these another :eyes: ?

ros2/nightly/platform.yaml
```diff
@@ -10,4 +10,4 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version:
```
ruffsl: I think the `version` parameter was previously used for setting the package version in the apt install. Leaving it empty was to signify to the template function to poll the apt repo for the latest version number. Did you refactor this logic in docker_templates to match your repurpose of it here? Given osrf apt repos don't archive past version for long, this repurpose of `version` you've made seems more useful.

ros/bouncy/ubuntu/bionic/platform.yaml
```diff
@@ -10,4 +10,4 @@ platform:
     maintainer_name:
     arch: amd64
     type: distribution
-    version:
+    version: 2
```
ruffsl: Could we update the ardent Dockerfiles with this? I know its EOL, but its helpful to keep the auto generate dockerfile in sync so that we only need the single `manifest` file to regenerate all folders in the library.

ros2/nightly/platform.yaml
mikaelarguedas: Oh I was actually wondering why some files had it https://github.com/osrf/docker_images/blob/8404682c98c1182203bcbe9898af6832f387aaab/ros2/source/platform.yaml#L12 and some didn't https://github.com/osrf/docker_images/blob/8404682c98c1182203bcbe9898af6832f387aaab/ros2/nightly/platform.yaml#L13

I didn't change any corresponding logic in the templates scripts. Surprising the generated files didnt change, maybe it's used only for gazebo? Do you remember where this is used exactly?

ros/bouncy/ubuntu/bionic/platform.yaml
mikaelarguedas: Sure thing. Do you want me to update the ROS1 EOL Dockerfiles as well?

ros2/nightly/platform.yaml
ruffsl: I think it was this:

https://github.com/osrf/docker_templates/blob/366392402917ff19b14fc0c5ed088bed8bd2e329/docker_templates/packages.py#L80

ros/bouncy/ubuntu/bionic/platform.yaml
ruffsl: If it's not to much trouble. Having updated `.config/ros1/platform.yaml.em` you should only need to delete the ros1 folders and run the `./create_dockerfolders.py` in the ros directory. You may need to toggle the comments of the EOL listings to ensure they get regenerated.

mikaelarguedas: Yeah that's no trouble at all. I'll update them all once we resolve the other pending question

ros2/nightly/platform.yaml
mikaelarguedas: Discussed this offline with @ruffsl, we will not use `version` but `ros_version` instead, that allows to not change the semantic and be consistent with the name of the environment variable `ROS_VERSION`

PR #246, opened 2019-03-22
Ros2 devel

mikaelarguedas:
Creates a `ros2:devel` image as discussed in https://github.com/osrf/docker_templates/pull/49

Depends on https://github.com/osrf/docker_templates/pull/54, https://github.com/osrf/docker_templates/pull/51 and https://github.com/osrf/docker_templates/pull/52

### DIFF
--- a/ros2/source/Makefile
+++ b/ros2/source/Makefile
@@ -10,10 +10,13 @@ help:
 	@echo ""
 
 build:
+	@docker build --tag=ros2:devel			devel/.
 	@docker build --tag=ros2:source			source/.
 
 pull:
+	@docker pull ros2:devel
 	@docker pull ros2:source
 
 clean:
+	@docker rmi -f ros2:devel
 	@docker rmi -f ros2:source
--- a/ros2/source/devel/Dockerfile
+++ b/ros2/source/devel/Dockerfile
@@ -1,0 +1,75 @@
+# This is an auto generated Dockerfile for ros2:devel
+# generated from docker_images_ros2/source/create_ros_image.Dockerfile.em
+FROM ubuntu:bionic
+
+# setup timezone
+RUN echo 'Etc/UTC' > /etc/timezone && \
+    ln -s /usr/share/zoneinfo/Etc/UTC /etc/localtime && \
+    apt-get update && apt-get install -q -y tzdata && rm -rf /var/lib/apt/lists/*
+
+# install packages
+RUN apt-get update && apt-get install -q -y \
+    bash-completion \
+    dirmngr \
+    gnupg2 \
+    lsb-release \
+    python3-pip \
+    wget \
+    && rm -rf /var/lib/apt/lists/*
+
+# setup keys
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 421C365BD9FF1F717815A3895523BAEEB01FA116
+
+# setup sources.list
+RUN echo "deb http://packages.ros.org/ros2/ubuntu `lsb_release -sc` main" > /etc/apt/sources.list.d/ros2-latest.list
+
+# setup environment
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
+
+# install bootstrap tools
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    git \
+    python3-colcon-common-extensions \
+    python3-rosdep \
+    python3-vcstool \
+    && rm -rf /var/lib/apt/lists/*
+
+# install python packages
+RUN pip3 install -U \
+    argcomplete \
+    flake8 \
+    flake8-blind-except \
+    flake8-builtins \
+    flake8-class-newline \
+    flake8-comprehensions \
+    flake8-deprecated \
+    flake8-docstrings \
+    flake8-import-order \
+    flake8-quotes \
+    pytest-repeat \
+    pytest-rerunfailures
+
+# bootstrap rosdep
+RUN rosdep init \
+    && rosdep update
+
+# clone source
+ENV ROS2_WS /opt/ros2_ws
+RUN mkdir -p $ROS2_WS/src
+WORKDIR $ROS2_WS
+
+# build source
+RUN colcon \
+    build \
+    --cmake-args -DSECURITY=ON --no-warn-unused-cli \
+    --symlink-install
+
+# setup bashrc
+RUN cp /etc/skel/.bashrc ~/
+
+# setup entrypoint
+COPY ./ros_entrypoint.sh /
+
+ENTRYPOINT ["/ros_entrypoint.sh"]
+CMD ["bash"]
--- a/ros2/source/devel/Dockerfile
+++ b/ros2/source/devel/Dockerfile
@@ -1,6 +1,7 @@
 # This is an auto generated Dockerfile for ros2:devel
-# generated from docker_images_ros2/source/create_ros_image.Dockerfile.em
-FROM ubuntu:bionic
+# generated from docker_images_ros2/devel/create_ros_image.Dockerfile.em
+ARG FROM_IMAGE=ubuntu:bionic
+FROM $FROM_IMAGE
 
 # setup timezone
 RUN echo 'Etc/UTC' > /etc/timezone && \
@@ -14,7 +15,6 @@ RUN apt-get update && apt-get install -q -y \
     gnupg2 \
     lsb-release \
     python3-pip \
-    wget \
     && rm -rf /var/lib/apt/lists/*
 
 # setup keys
@@ -50,8 +50,6 @@ RUN pip3 install -U \
     pytest-repeat \
     pytest-rerunfailures
 
-ENV ROS_DISTRO crystal
-
 # bootstrap rosdep
 RUN rosdep init \
     && rosdep update
@@ -64,7 +62,8 @@ WORKDIR $ROS2_WS
 # build source
 RUN colcon \
     build \
-    --cmake-args -DSECURITY=ON --no-warn-unused-cli \
+    --cmake-args \
+      -DSECURITY=ON --no-warn-unused-cli \
     --symlink-install
 
 # setup bashrc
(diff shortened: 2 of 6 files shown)


ros2/source/Makefile
```diff
@@ -10,10 +10,13 @@ help:
 	@echo ""
 
 build:
+	@docker build --tag=ros2:devel			devel/.
```
ruffsl: Would you want the `source` image to build FROM the `devel`? This Could reduce duplication.

mikaelarguedas: yeah that was my original idea. my understanding from https://github.com/osrf/docker_templates/pull/49#issuecomment-474635769 was that you liked the idea of using the same template from both.

But we can go the other way and reduce the source template to be based on devel and just clone, build and source the workspace

ruffsl: Using nested tags may complicate automated builds given the pricipaled ordering of builds, but given we aren't doing that with `source`, and only the `nightly` tag is built via osrf's automated repo on Dockerhub, I think we are fine in separating the devel setup and the source compilation into separate Dockerfiles.

ros2/source/images.yaml.em
```diff
@@ -47,3 +42,34 @@ images:
         vcs:
             ros2:
                 repos:
+
+    devel:
```
ruffsl: Could we reorder `source, devel` entries in the yaml to `devel, source` to mirror there depency ording?  
This is done with all the other tag configs, e.g. ros-core -> ros-base -> ...

ros2/source/devel/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+    flake8-class-newline \
+    flake8-comprehensions \
+    flake8-deprecated \
+    flake8-docstrings \
+    flake8-import-order \
+    flake8-quotes \
+    pytest-repeat \
+    pytest-rerunfailures
+
+ENV ROS_DISTRO crystal
```
ruffsl: As we spoke offline, shall we move `ENV ROS_DISTRO` to source?

mikaelarguedas: I've been working on a refactor based on offline discussion, on this branchhttps://github.com/osrf/docker_templates/tree/ros2_devel_reboot, it will provide arg for ros distro in devel as well as args for testing ( and fail on test failure) in the source template. Not ready for review yet, I'll submit something in the coming days with the changes

ros2/source/Makefile
mikaelarguedas: :+1: please see https://github.com/osrf/docker_templates/pull/57 for a version with nested templates

ros2/source/devel/Dockerfile
mikaelarguedas: https://github.com/osrf/docker_templates/pull/57 now has a distro agnostic `devel` and allow users to set `ROS_DISTRO` on the source image using build arguments

ros2/source/source/ros_entrypoint.sh
```diff
@@ -2,5 +2,5 @@
 set -e
```
ruffsl: This `ros_entrypoint.sh` file is no longer used by the `source` Dockerfile, as well as made redundant given the parent image, `devel`, already includes an approrape entrypoint.

ros2/source/source/Dockerfile
```diff
@@ -72,17 +27,20 @@ RUN apt-get update && rosdep install -y \
...
+        colcon test; \
+        if [ ! -z "$FAIL_ON_TEST_FAILURE" ]; then \
+            colcon test-result; \
+        else \
+            colcon test-result || true; \
+        fi \
+    fi
 
-ENTRYPOINT ["/ros_entrypoint.sh"]
 CMD ["bash"]
```
ruffsl: Setting of the command in `source` is also redundant given its set in the `devel` image.

ros2/source/source/Dockerfile
```diff
@@ -72,17 +27,20 @@ RUN apt-get update && rosdep install -y \
...
     --symlink-install
 
-# setup bashrc
-RUN cp /etc/skel/.bashrc ~/
-
-# setup entrypoint
-COPY ./ros_entrypoint.sh /
+ARG RUN_TESTS
+ARG FAIL_ON_TEST_FAILURE
+RUN if [ ! -z "$RUN_TESTS" ]; then \
```
ruffsl: Awesome!

ros2/source/source/Dockerfile
```diff
@@ -72,17 +27,20 @@ RUN apt-get update && rosdep install -y \
     && rm -rf /var/lib/apt/lists/*
 
 # build source
-WORKDIR $ROS2_WS
 RUN colcon \
     build \
     --cmake-args -DSECURITY=ON --no-warn-unused-cli \
```
ruffsl: Could we tabulate these so colcon build args are revisioned on separate lines for VCS?
```suggestion
     --cmake-args \
       -DSECURITY=ON \
     --no-warn-unused-cli \
```

ros2/source/images.yaml.em
```diff
@@ -2,19 +2,15 @@
 # ROS2 Dockerfile database
 ---
 images:
-    source:
+    devel:
```
ruffsl: Thanks for recording devel and source.

ros2/source/devel/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+# clone source
+ENV ROS2_WS /opt/ros2_ws
+RUN mkdir -p $ROS2_WS/src
+WORKDIR $ROS2_WS
+
+# build source
+RUN colcon \
+    build \
+    --cmake-args \
+      -DSECURITY=ON --no-warn-unused-cli \
```
ruffsl:
```suggestion
      -DSECURITY=ON \
    --no-warn-unused-cli \
```

ros2/source/source/ros_entrypoint.sh
mikaelarguedas: :+1: good catch, this is a leftover, it is not used by the new template. I deleted the source and devel folders to make sure to remove any deprecated files in https://github.com/osrf/docker_images/pull/246/commits/897cf8408b46c0edf5ce0781c4b4644cb6f7fd82

ros2/source/source/Dockerfile
mikaelarguedas: This comes from the images.yaml.em file. This PR doesn't modify the logic surrounding colcon args expansion.
Trying to keep the scope here minimal I was not planning on addressing this in this PR.

ros2/source/devel/Dockerfile
mikaelarguedas: re https://github.com/osrf/docker_images/pull/246#discussion_r268835249

ros2/source/source/Dockerfile
mikaelarguedas: Good point.
I originally kept the template logic in case we wanted to provide more commands to be set here.
I don't see a strong use case for it though so I'll remove it completely from the template.

Does that sound good?

mikaelarguedas: removed in https://github.com/osrf/docker_templates/pull/57/commits/d9e085206f4c66e661cdd314c6675df371fbf6bc resulting in https://github.com/osrf/docker_images/pull/246/commits/b4924e918e81524c4e05570aa91cec1495cdd247

ros2/source/source/Dockerfile
ruffsl: :lipstick: later then.

ros2/source/devel/Dockerfile
ruffsl: :lipstick: later then.

mikaelarguedas:
Note: for this to work, we will need to push to the osrf dockerhub the `ros2:devel` image.
`nightly` seems to be built by a cron job and source is never pushed to dockerhub.
I dont think devel would change often, is there a preferred strategy for pushing `devel` to dockerhub ?

ros2/source/source/Dockerfile
```diff
@@ -72,17 +27,18 @@ RUN apt-get update && rosdep install -y \
     && rm -rf /var/lib/apt/lists/*
 
 # build source
-WORKDIR $ROS2_WS
 RUN colcon \
```
ruffsl: Oh wait, are we missing a rosdep install here before building?

mikaelarguedas: The previous `RUN` statement (L23) is a `rosdep install`. Is that what you are referring to ?

ruffsl:
> I dont think devel would change often, is there a preferred strategy for pushing devel to dockerhub ?

I think it may be worthwhile to include `devel` into the cron job OSRF has that calls our Docker Hub webhook, currently only triggering the `nightly` tag. Having the `devel` rebuilt every day to keep fresh may be helpful to catch upstream package changes in ubuntu, etc. I think we could leave out `source` tag from osrf/ros2 repo given it make more sense for users to build locally, and may be more likely to fail in building.

mikaelarguedas:
> Having the devel rebuilt every day to keep fresh may be helpful to catch upstream package changes in ubuntu, etc. I think we could leave out source tag from osrf/ros2 repo given it make more sense for users to build locally, and may be more likely to fail in building.

Sounds good to me :+1: 

> worthwhile to include devel into the cron job OSRF has that calls our Docker Hub webhook

@nuclearsandwich is this something that you'd be ok with? (assuming you're the person to ask :) )

ros2/source/source/Dockerfile
ruffsl: nvm, github collapsed those lines in the diff preview when I was skimming the latest changes.

mikaelarguedas:
@nuclearsandwich very gentle ping, just to make sure the previous notification didnt get lost :)
Feel free to ping me offline if there are details you'd like to iron out

nuclearsandwich:
> @nuclearsandwich is this something that you'd be ok with? (assuming you're the person to ask :) )

Yep. As of now this cron job is just running on my workstation.

ruffsl:
> Yep. As of now this cron job is just running on my workstation.

Cool, I'm going to merge and test the docker hub build of `devel` via the admin console.

PR #247, opened 2019-03-22
Update ros2:nightly image

mikaelarguedas:
Relevant changes:
- add missing tinydir_vendor rosdep rule: https://github.com/osrf/docker_images/pull/247/commits/7afe66362c533572a6be58ecd64cf63584b1cc4c
- updated nightly image config https://github.com/osrf/docker_images/pull/247/commits/ecf8e65640436dd99403a44ffe0cd77426df505a
- updated ROS2 nightly Dockerfile https://github.com/osrf/docker_images/pull/247/commits/9210d0d9992df92f34b6b0450f803521b5483789

Dockerfile generated using https://github.com/osrf/docker_templates/pull/55

### DIFF
--- a/ros2/nightly/nightly/prereqs.yaml
+++ b/ros2/nightly/nightly/prereqs.yaml
@@ -2,8 +2,4 @@ console_bridge: {ubuntu: []}
 fastcdr: {ubuntu: []}
 fastrtps: {ubuntu: []}
 osrf_testing_tools_cpp: {ubuntu: []}
-poco_vendor: {ubuntu: []}
-tinyxml_vendor: {ubuntu: []}
-tinyxml2_vendor: {ubuntu: []}
-urdfdom: {ubuntu: []}
 urdfdom_headers: {ubuntu: []}
--- a/ros2/nightly/nightly/prereqs.yaml
+++ b/ros2/nightly/nightly/prereqs.yaml
@@ -2,4 +2,5 @@ console_bridge: {ubuntu: []}
 fastcdr: {ubuntu: []}
 fastrtps: {ubuntu: []}
 osrf_testing_tools_cpp: {ubuntu: []}
+tinydir_vendor: {ubuntu: []}
 urdfdom_headers: {ubuntu: []}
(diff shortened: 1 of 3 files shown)


ros2/nightly/nightly/prereqs.yaml
```diff
@@ -2,8 +2,5 @@ console_bridge: {ubuntu: []}
 fastcdr: {ubuntu: []}
 fastrtps: {ubuntu: []}
 osrf_testing_tools_cpp: {ubuntu: []}
-poco_vendor: {ubuntu: []}
-tinyxml_vendor: {ubuntu: []}
-tinyxml2_vendor: {ubuntu: []}
-urdfdom: {ubuntu: []}
+tinydir_vendor: {ubuntu: []}
```
ruffsl: Do we still need this if we are installing `ros-$ROS_DISTRO-ros-workspace`?

mikaelarguedas: yes. It is the other way around, we can afford to add this line now that we install `ros-$ROS_DISTRO-ros-workspace`.

---

Once #245 merged, there should be no `ros-crystal-*` packages installed in the nightly image except from `tinydir_vendor` and `ros_workspace`.

With the nightly PR installing `ros-$ROS_DISTRO-ros-workspace`, we do not need to install any other ros-crystal package from debs. So we can stop installing that one remaining package.

As of why this package needs to be listed, my guess it that it's a pure cmake package and not an ament package, so it doesnt install it's package.xml nor register itself in the ament index.

nuclearsandwich: @cottsay contributed https://github.com/ros2/tinyxml2_vendor/pull/6 which resolves this issue for tinyxml2_vendor. Something similar could be done for tinydir_vendor.

mikaelarguedas: :+1: that makes sense for vendor packages as it doesn't really matter if they're not registered in the ament index.

PR #255, opened 2019-03-30
source setup files before calling rosdep install

mikaelarguedas:
This updates the nightly image config according to https://github.com/osrf/docker_templates/pull/58

### DIFF
--- a/ros2/nightly/images.yaml.em
+++ b/ros2/nightly/images.yaml.em
@@ -27,8 +27,8 @@ images:
         rosdep_override:
             - prereqs.yaml
         rosdep:
-            install:
-                - --from-paths /opt/ros/$ROS_DISTRO/share
-                - --ignore-src
-                - --skip-keys "libopensplice69 rti-connext-dds-5.3.1"
+            path: /opt/ros/$ROS_DISTRO/share
+            skip_keys:
+                - libopensplice69
+                - rti-connext-dds-5.3.1
         ros2_binary_url: https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -61,6 +61,11 @@ ARG ROS2_BINARY_URL=https://ci.ros2.org/view/packaging/job/packaging_linux/lastS
 RUN wget -q $ROS2_BINARY_URL -O - | \
     tar -xj --strip-components=1 -C /opt/ros/$ROS_DISTRO
 
+# install setup files
+RUN apt-get update && apt-get install -q -y \
+    ros-$ROS_DISTRO-ros-workspace \
+    && rm -rf /var/lib/apt/lists/*
+
 # add custom rosdep rule files
 COPY prereqs.yaml /etc/ros/rosdep/
 RUN echo "yaml file:///etc/ros/rosdep/prereqs.yaml" | \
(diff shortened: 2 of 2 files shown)


ruffsl:
Should we do the same for the `source` image?

https://github.com/osrf/docker_images/blob/99cf9528aa57b3e45a8ca1e534212469557b08d3/ros2/source/source/Dockerfile#L22-L27

mikaelarguedas:
> Should we do the same for the source image?
Sourcing the workspace from devel that is. $ROS2_WS/install/setup.bash

The devel setup file script will not set any of these env variable as the devel image is distro agnostic:
```
$ docker run -it --rm osrf/ros2:devel
root@07ccdf1b3aea:/opt/ros2_ws# source install/setup.bash 
root@07ccdf1b3aea:/opt/ros2_ws# env | grep ROS
ROS2_WS=/opt/ros2_ws
```

Options we have:
1. include the [ros_environment package](https://github.com/ros/ros_environment) in the workspace
1. install the `ros_enviroment` deb
1. set the env vars `ROS_PYTHON_VERSION` and `ROS_VERSION` in the dockerfile

1 and 2, if done in the devel image, will make the image distro specific, so it would be better to do it in the source image.
3 duplicates code that's already in the ros_environment package, but keeps the devel image distro-agnostic while being ros2 specific.

mikaelarguedas:
> Shouldn't this be included in the ros2.repo file? Is there a reason its currently not?

Sorry I was originally thinking of the devel image and meant "place it in the underlay workspace" (aka a workspace we can source before building the source workspace)
it is included in the ros2.repos file: https://github.com/ros2/ros2/blob/3c43c6b839dcdac6138037e402f28e8aa556de4c/ros2.repos#L110
But the environment hooks won't be invoked until the workspace is built. So 1 is not a solution if we target changes to the source image.

> We'd have to be sure to source /opt/ros/$ROS_DISTRO before creating the $ROS2_WS workspace, no?

Yes we would need to source it before invoking `rosdep install`

> This would further depart from the president of ENV set in the official ros images. Though I suppose that is more of a objective for nightly, not necessary currently for source.

Yes, though any consumer of the docker image using the ros entrypoint would have all of these set. They're just not explicitly set in the dockerfile because it would be duplicating what is set by the setup files.
```
$ docker run -it --rm ros:crystal
root@9f894752c09b:/# env | grep ROS
ROS_VERSION=2
ROS_PYTHON_VERSION=3
ROS_DISTRO=crystal
```

So I don't feel too bad about adding those to the Dockerfile. I actually find it less confusing than having an arbitrary ros crystal deb installed. But I'm fine either way

ruffsl:
> But the environment hooks won't be invoked until the workspace is built. So 1 is not a solution if we target changes to the source image.

How stable is the `ros_enviroment` package? Is it likely to change?

> So I don't feel too bad about adding those to the Dockerfile. I actually find it less confusing than having an arbitrary ros crystal deb installed. But I'm fine either way

The setting of `ROS_VERSION=2`  and `ROS_PYTHON_VERSION=3` are probably not likely to change soon for ros2, so setting them in the Dockerfile would be simple. I'm just cautious if we'll have to keep replicating more and more of `ros_enviroment` behavior, e.g. if we start having to track ENV for expected C, Rust, Go versions etc.

mikaelarguedas:
> How stable is the ros_enviroment package? Is it likely to change?

Pretty unlikely.
There may be `ROS_ETC_DIR` set in the future like in ROS 1 but it's about the only difference between the ROS1 and ROS2 versions of the package ATM.

> I'm just cautious if we'll have to keep replicating more and more of ros_enviroment behavior

Yeah I agree with that, not a big fan of duplicating information here.

> e.g. if we start having to track ENV for expected C, Rust, Go versions etc.

I would not expect that to happen, if it does in a way tat impacts package.xml writing and rosdep resolution it will be mentionned in a REP. e.g. the rationale for `ROS_VERSION` and `ROS_PYTHON_VERSION` is part of the [package format 3 rep](https://github.com/ros-infrastructure/rep/blob/master/rep-0149.rst#environment-variables)

ruffsl:
Well, as this connor case is unlikely to change soon, setting them ENV seems simple:

``` Dockerfile
ARG ROS_DISTRO=crystal
ENV ROS_DISTRO=$ROS_DISTRO
ENV ROS_PYTHON_VERSION=3 \
    ROS_VERSION=2
```

If it starts changing later, perhaps switching to `ros_enviroment` via underlay workspace would be worth it.

PR #261, opened 2019-05-06
update ros2:nightly for dashing

mikaelarguedas:
the `ros_environment` package [now points to `dashing`](https://github.com/ros2/ros2/pull/688) making the value of the environment variable inconsistent in the dockerfile between before and after sourcing the extracted workspace. This causes the nightly image to fail to build for the last week.

This fix needs the `ros-dashing-ros-workspace` deb to be available. I changed the nightly template to use the `ros2-testing` apt repository to get the package, this seems an acceptable workaround for the nightly image to use the latest released packages rather than the ones synced to main. If we prefer to stick to main, this PR needs to wait for the first sync to main of Dashing beta before being merged.

@ruffsl @nuclearsandwich do you know if there are notifications emails for the failing nightly builds ?

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -55,7 +55,7 @@ RUN pip3 install -U \
     pytest-rerunfailures
 
 # install ros2 packages
-ENV ROS_DISTRO crystal
+ENV ROS_DISTRO dashing
 RUN mkdir -p /opt/ros/$ROS_DISTRO
 ARG ROS2_BINARY_URL=https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
 RUN wget -q $ROS2_BINARY_URL -O - | \
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -21,7 +21,7 @@ RUN apt-get update && apt-get install -q -y \
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 421C365BD9FF1F717815A3895523BAEEB01FA116
 
 # setup sources.list
-RUN echo "deb http://packages.ros.org/ros2/ubuntu `lsb_release -sc` main" > /etc/apt/sources.list.d/ros2-latest.list
+RUN echo "deb http://packages.ros.org/ros2-testing/ubuntu `lsb_release -sc` main" > /etc/apt/sources.list.d/ros2-testing.list
 
 # install bootstrap tools
 RUN apt-get update && apt-get install --no-install-recommends -y \
--- a/ros2/nightly/platform.yaml
+++ b/ros2/nightly/platform.yaml
@@ -5,7 +5,7 @@ platform:
     os_name: ubuntu
     os_code_name: bionic
     rosdistro_name: melodic
-    ros2distro_name: crystal
+    ros2distro_name: dashing
     user_name: ros
     maintainer_name:
     arch: amd64


nuclearsandwich:
> do you know if there are notifications emails for the failing nightly builds ?

I certainly don't get one and a quick look through https://docs.docker.com/docker-hub/builds/ doesn't indicate that notification emails are an available feature.

mikaelarguedas:
> I certainly don't get one and a quick look through https://docs.docker.com/docker-hub/builds/ doesn't indicate that notification emails are an available feature.

Thanks ! I guess a way to do it would be to extend travis cron job to build the ros2 images but not build them on PR or commits to the master branch. This way we would receive emails on failure but not lengthen CI time.

mikaelarguedas:
There seem to be a "Notifications" sections in the docker hub settings section, that are right now on "off". We could set it to "Only failures" instead.

This may be a bit noisy though as `Notification emails for organizations are sent to the members of the owners team.` so this means that all the members of that team will receive emails on build failures of any image in the org.

ruffsl:
> I guess a way to do it would be to extend travis cron job to build the ros2 images

I'd prefer to keep the image build jobs delegated to Docker Hub rather than another CI, and we already rely bit on all the DockerHub/GitHub integrations and webhooks. I'm not sure I'd like to replicate that scafulting with more scripts here. Perhaps owners can add an email inbox rule/filter. I already have number for GitHub orgs and CI tools.

> this seems an acceptable workaround for the nightly image to use the latest released packages rather than the ones synced to main. If we prefer to stick to main, this PR needs to wait for the first sync to main of Dashing beta before being merged.

Would going with `ros2-testing` be suitable for downstream CI? As that's what `nightly` is presently being used for. I'd be in favor of restoring `nightly` builds ASAP though.

---

In the future, would there be a better approach for transitioning to future distros? Like sticking with the `master` dev repo for a particular distro until the next distro's nightly is ready. I guess there's not necessarily a nightly fat archives particular to distro (too much of an oxymoron)?

mikaelarguedas:
> Would going with ros2-testing be suitable for downstream CI? As that's what nightly is presently being used for. I'd be in favor of restoring nightly builds ASAP though.

The places where I use `ros2:nightly` for CI currently don't rely on any released debs of a particular distro. I would be of the opinion that users/repos tracking the nightly builds should *not* expect debs for a given distro to be API compatible and build from source any package not provided in the archive (if they want the debs of a specific distro, using `ros:<DISTRO>` as the base image would be more appropriate). At that point, I would argue that `ros2` or `ros2-testing` doesn't make much of a difference.
Does your experience working with projects relying on `ros2:nightly` for CI suggests an apt repo would be better than the other?

---

> In the future, would there be a better approach for transitioning to future distros? Like sticking with the master dev repo for a particular distro until the next distro's is ready. I guess there's not necessarily a nightly fat archives particular to distro (too much of an oxymoron)?

For context, we started discussing this on https://github.com/osrf/docker_images/pull/245#issuecomment-475085192 where my suggestion was to transition in the alpha cycle of the distro as soon as `ros-workspace` deb is available

I guess we need to break down what is distro specific in the master/nightly image to figure out the best course of action:
- In the archive itself: The only "distro-specific" thing is the package populating the `ROS_DISTRO` env var (`ros_environment`).
- In the nightly image:
  - The `ROS_DISTRO` env var being set in the Dockerfile allowing to
  - install the `ros_workspace` package installed from debs to provide the setup files

Assuming we want / need to keep both these things as they are, what seems to be a good chronology for transitioning docker images for new distros would be the following:

- Alpha release of the distro:
  - release of `ros-workspace` deb into the new distro (+ sync to main repo?)
  - change `ros2.repos` to point to the new branch of `ros_environment`
  - update nightly image to switch to the new distro.

- Beta release of a distro:
  - sync to main of the packages for wide user testing
  - update the source image to use the tagged `release-<DISTRO>` repos file (if not tagged until official release, move this item to next section)

- Official release:
  - create the official docker images for the new distro
  - submit them to the official library

@nuclearsandwich insights on the timeline would be valuable as he has experience with ROS2/OSRF release timelines.

nuclearsandwich:
> Would going with `ros2-testing` be suitable for downstream CI?

In some ways the ros2-testing repo is very suitable for CI as it is used by dev and PR jobs on the official build farm; so using it brings external CI closer to parity with what would be tested on the official farm.

The testing repositories are less stable and changes from the fresh "building" repository are synchronized to testing as soon as there are no other queued rebuilds for a platform. There's [a threshold setting](https://github.com/ros2/ros_buildfarm_config/blob/08c59c6d8680b5cf25f162172d547d1a9db6f665/crystal/release-build.yaml#L19) to prevent a bad build run from nuking every package in testing but if an external CI relies on a high level package it's possible that a package could go temporarily missing due to a build failure.

> In the future, would there be a better approach for transitioning to future distros?

There's a bit of a chicken and egg here since it's expected that most consumers of the ROS_DISTRO name will get it from the ros_environment package, but in this case the name needs to be known so the package can be sourced..

In this case the breakage was caused by https://github.com/ros2/ros2/pull/688 which is where we updated the ros_environment for the master repos file (which feeds the packaging builds) to use dashing as master was already far-diverged from crystal. 

I'm also generally confused as to what's coming from the ros_workspace deb package for nightly builds.

---

The timeline proposed by @mikaelarguedas above seems sound. When the API freeze settles and bloom releases are made I'll sync the results to the main repo as well.

mikaelarguedas:
> I'm also generally confused as to what's coming from the ros_workspace deb package for nightly builds.

Some additional context can be found in https://github.com/osrf/docker_images/pull/245#issuecomment-475070705.

In a nutshell: the setup files provided by the nightly archive rely on `COLCON_PREFIX_PATH` being set manually (otherwise defaulting to `/home/jenkins-agent/workspace/packaging_linux/ws/install`) and a chained melodic workspace causing error messages for users when sourcing.
Using ros-workspace works around the issue by using the colcon and ROS1 agnostic setup files that users use when they install ROS2 from debs.

ruffsl:
> In some ways the ros2-testing repo is very suitable for CI as it is used by dev and PR jobs on the official build farm; so using it brings external CI closer to parity with what would be tested on the official farm.

:+1: 

> The timeline proposed by @mikaelarguedas above seems sound.

Agreed. Thanks @mikaelarguedas for breaking that down.

PR #267, opened 2019-06-01
Adding ROS2 Dashing release

ruffsl:
Context: https://discourse.ros.org/t/ros-2-dashing-diademata-released/9365

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install:
 
 env:
   # ros2
+  - HUB_REPO=ros   HUB_RELEASE=dashing  HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=bionic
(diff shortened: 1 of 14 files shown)


mikaelarguedas:
thank @ruffsl for adding dashing, we may want to wait for https://github.com/ros2-gbp/rmw_implementation-release/pull/2 to be synced to avoid people relying on the images containing opensplice for size + behavior reasons (this is currently breaking CI on some of my repos with opensplice message generation failing)

ruffsl:
> we may want to wait for ros2-gbp/rmw_implementation-release#2 to be synced to avoid people relying on the images containing opensplice for size + behavior reasons

Oh, I was not aware of this regression. Definitely hope that lands soon.

PR #269, opened 2019-06-02
[WIP] Rotate keys for ros1 apt repos

ruffsl:


tfoote:
Note that the EOL distros will need a new url and different key with the [planned migration](https://discourse.ros.org/t/security-issue-on-ros-build-farm/9342/8?u=tfoote)

ruffsl:
@tfoote  , @nuclearsandwich Ok, so if read that correctly, these would be the respective changes?:

### For Indigo, Kinetic, Lunar, and Melodic

``` diff
- 421C365BD9FF1F717815A3895523BAEEB01FA116
+ C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
```

### For older/other unsupported distros

``` diff
- packages.ros.org
+ snapshots.ros.org
```

nuclearsandwich:
I have a commit to make on top of this that does the right thing for snapshots. It will have to wait for tomorrow though as I can't keep my eyes open any longer.

nuclearsandwich:
I opened https://github.com/osrf/docker_templates/pull/64 which I believe sits upstream of this change and should produce the desired results once the snapshot repos are live later today.

eisoku9618:
> @tfoote , @nuclearsandwich Ok, so if read that correctly, these would be the respective changes?:

I create a PR (https://github.com/osrf/docker_images/pull/278) for the changes. Please merge the PR (https://github.com/osrf/docker_images/pull/278) to this PR for the EOL distros.

eisoku9618:
I use EOL distro docker images, so I would appreciate if you could update EOL distro tags of Dockerhub.

mikaelarguedas:
Superseded by https://github.com/osrf/docker_images/pull/296 based on the templates from https://github.com/osrf/docker_templates/pull/67

Issue #270, opened 2019-06-04
ros2:source fails to build because pytest cant be found


mikaelarguedas:
Currently the source image fails to build for both crystal and dashing. Despite `python3-pytest` being installed on the system, the module cannot be found by `ament_cmake_pytest` resulting in warnings like:
```
--- stderr: rosidl_adapter
CMake Warning at /opt/ros2_ws/install/ament_cmake_pytest/share/ament_cmake_pytest/cmake/ament_add_pytest_test.cmake:73 (message):
  The Python module 'pytest' was not found, pytests can not be run (e.g.  on
  Ubuntu/Debian install the package 'python3-pytest')
Call Stack (most recent call first):
  CMakeLists.txt:31 (ament_add_pytest_test)
```

Some pytests targets in the ROS stack are not checked for existence before setting properties leading to CMake Errors.

Needs some more investigation to figure out why the module is not found and the list of tests that should check for target existence

clalancette:
This looks to be a generic problem with the source install, as @Karsten1987 ran into the same problem and I reproduced it earlier today.

mikaelarguedas:
Yeah working with @Karsten1987 yesterday we ran into it in multiple places.

Is there an issue tracking it on the ROS2 repositories? I tracked down the issue to the pytest version and can suggest pinning the version for ROS2 users until it is fixed upstream or a better workaround is found

clalancette:
> Is there an issue tracking it on the ROS2 repositories? I tracked down the issue to the pytest version and can suggest pinning the version for ROS2 users until it is fixed upstream or a better workaround is found

Not at the moment.  If you don't mind opening an issue and suggesting a patch, that would be great!  Thanks.

mikaelarguedas:
For reference: this issue is being tracked at https://github.com/ros2/ros2_documentation/issues/271 instead of https://github.com/ros2/ros2/issues/722

PR #272, opened 2019-06-04
[ros2:devel / ros2:nightly] install colcon from pip and not apt

mikaelarguedas:
Related to https://github.com/ros2/ros2/issues/722

Fixes https://github.com/osrf/docker_images/issues/270

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -26,7 +26,6 @@ RUN echo "deb http://packages.ros.org/ros2-testing/ubuntu `lsb_release -sc` main
 # install bootstrap tools
 RUN apt-get update && apt-get install --no-install-recommends -y \
     git \
-    python3-colcon-common-extensions \
     python3-rosdep \
     python3-vcstool \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


ruffsl:
Would this PR be reverted once the packaging issue with `pytest` is resolved? I recall we previously migrated away from installing colcon from pip to apt for downstream consistency of dev tools.

Related: https://github.com/pytest-dev/pytest/pull/5389

mikaelarguedas:
> Would this PR be reverted once the packaging issue with pytest is resolved?

Originally the goal is to install all the third-party tools from APT and not pip to allow conflict tracking and automatic upgrade.
In this particular case, I'm not convinced it needs to be reverted.
As this PR targets only the images that are expected to be built daily it shouldn't be too big of a deal to install colcon from pip. It also allows to move away from the error prone pattern of installing some pytest plugins from pip and the others from APT.
I doubt the fix will ever make it to bionic so we will need to install pytest via pip to get it.

ruffsl:
> I doubt the fix will ever make it to bionic so we will need to install pytest via pip to get it.

Ah, that would do it.
Does this mean that `library/ros:dashing` images are left with a broken build tool?

mikaelarguedas:
> Does this mean that library/ros:dashing images are left with a broken build tool?

No they should be fine as they dont install any plugins from pip. They just are not able to leverage the various pytest options that colcon provides (I think, I haven't built or tried the dashing images)

PR #273, opened 2019-06-05
Update Dockerfiles for ROS repository changes.

nuclearsandwich:
This replaces #269 and is generated from the union of https://github.com/osrf/docker_templates/pull/63 and https://github.com/osrf/docker_templates/pull/64

### DIFF
--- a/ros/kinetic/debian/jessie/ros-core/Dockerfile
+++ b/ros/kinetic/debian/jessie/ros-core/Dockerfile
@@ -1,43 +1,29 @@
 # This is an auto generated Dockerfile for ros:ros-core
-# generated from docker_images/create_ros_core_image.Dockerfile.em
+# generated from docker_images_legacy/create_ros_core_image.Dockerfile.em
 FROM debian:jessie
 
-# install packages
-RUN apt-get update && apt-get install -q -y \
-    dirmngr \
-    gnupg2 \
-    lsb-release \
-    && rm -rf /var/lib/apt/lists/*
+# setup source.list to old-releases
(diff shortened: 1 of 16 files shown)


ros/kinetic/debian/jessie/ros-core/Dockerfile
```diff
@@ -1,43 +1,29 @@
...
 FROM debian:jessie
 
-# install packages
-RUN apt-get update && apt-get install -q -y \
-    dirmngr \
-    gnupg2 \
-    lsb-release \
-    && rm -rf /var/lib/apt/lists/*
+# setup source.list to old-releases
+RUN sed -i -e 's/archive/old-releases/g' /etc/apt/sources.list
```
mikaelarguedas: There is something risky in applying this sed unconditionnally, e.g. this is meant for ubuntu, the regex should not match on jessie but ideally this run statement would not be present at all.

As we also don't know when e.g. trusty will transition to the `old-releases` repo, I suggest leveraging the list maintained in the `ros_buildfarm` repo. See https://github.com/osrf/docker_templates/pull/65 for a potential fix.

nuclearsandwich: Thanks @mikaelarguedas. @ruffsl feel free to merge that PR or do whatever you need to to this one. I will be unreachable for a few days in like an hour.

mikaelarguedas:
Superseded by https://github.com/osrf/docker_images/pull/296 based on the templates from https://github.com/osrf/docker_templates/pull/67

PR #274, opened 2019-06-05
Work around rviz_common import error

ruffsl:
Related:
https://github.com/ament/ament_cmake/issues/173
https://github.com/ros-planning/navigation2/issues/799
https://github.com/ros-planning/navigation2/pull/806

Ping @crdelsey

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -92,6 +92,11 @@ RUN . /opt/ros/$ROS_DISTRO/setup.sh \
 # ignore installed rosdep keys
 ENV ROS_PACKAGE_PATH /opt/ros/$ROS_DISTRO/share
 
+# FIXME Remove this once ament_export_interfaces respects COLCON_CURRENT_PREFIX https://github.com/ament/ament_cmake/issues/173
+#Workaround hard coded paths in nightly tarball setup scripts
+ARG UPSTREAM_CI_WS=/home/jenkins-agent/workspace/packaging_linux/ws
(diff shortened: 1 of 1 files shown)


crdelsey:
Thanks. I'll hold off on merging my workaround since you got this done so fast.

ruffsl:
@osrf/docker-image-maintainers , feel free to merge if approved.

mikaelarguedas:
Yeah that's one way to work around it.

I don't see a working case before that would be broken by this change and as it seems to fix navigation2 ci that's fine by me as a temporary workaround :+1: .

From the issue description it sounds like the issue is assumed to be in the nightly image, but at first glance it looks like any user using the extracted archive as an underlay will be facing the same problem (docker involved or not), is that correct?

@crdelsey Will the end issue and long term fix be located on the ros2/rviz repository ?

crdelsey:
> any user using the extracted archive as an underlay will be facing the same problem (docker involved or not), is that correct

Yes that is correct.

> Will the end issue and long term fix be located on the ros2/rviz repository ?

I don't know. I don't think so.  I don't understand the correct solution yet. There are 3 possibilities I can see:
1. Fix ament. Since CMake doesn't seem to support what we need here, it probably means making a custom equivalent of what CMake does internally.
2. Change the rviz build to not use interfaces. I'm not sure they are really necessary in this case, but there are other packages that use them as well, so they'd still have the problem
3. Change the way the build farm does the nightly build, so that the packages end up in the /opt/ros/... directory tree.

mikaelarguedas:
thanks @crdelsey for the clarification :+1:

Issue #283, opened 2019-06-20
Difference between these images and the ROS Docker Official Images is unclear


NoahAndrews:
My understanding is that most users (especially of ROS2) will want to use the Docker Official Images located at https://hub.docker.com/_/ros, so shouldn't the README (and by extension the [`osrf/ros2` Docker Hub page](https://hub.docker.com/r/osrf/ros2/)) link to that page? I almost went through the trouble of maintaining my own ROS2 dockerfile, because the one here doesn't support ARM, and I didn't realize that the Docker Official images had builds for ROS2.

mikaelarguedas:
Thanks @NoahAndrews for pointing out that the Readme was outdated.

See https://github.com/osrf/docker_images/pull/285 for a minimal change trying to clarify things

NoahAndrews:
👍

mikaelarguedas:
@NoahAndrews can you please check if https://github.com/osrf/docker_images/pull/305 addresses this issue?

NoahAndrews:
That's perfect!

NoahAndrews:
Thanks @ruffsl

PR #285, opened 2019-06-20
ROS 2 images are also hosted in the official ros library

mikaelarguedas:
Minimal change to point ROS 2 users to the official library.
We could consider providing more context about what "ROS 2 development images" mean

Attempt to fix https://github.com/osrf/docker_images/issues/283


mikaelarguedas:
Superseded by https://github.com/osrf/docker_images/pull/305

@NoahAndrews FYI

PR #301, opened 2019-08-02
Remove EOL distros from manifest

mikaelarguedas:
Replaces https://github.com/osrf/docker_images/pull/268/files
Fixes https://github.com/osrf/docker_images/issues/282

Should wait for all final images to be rebuilt on dockerhub before submitting upstream.

@nuclearsandwich Is Bouncy officially EOL?

PR #302, opened 2019-08-06
dashing patch 2

mikaelarguedas:
Did not update the docker library as the `ros1_bridge` image is not hosted on the official library

Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>


mikaelarguedas:
Yeah this has been synced as I noticed this when [CI failed](https://travis-ci.org/osrf/docker_images/jobs/568100247) on #301 because of it

PR #312, opened 2019-08-27
Use cron jobs from Circleci to trigger Docker Hub builds

ruffsl:
It seems like the server hosting the cron jobs for triggering Docker Hub builds keeps going AWOL.
I'd like to add some reliability and transparency to this task so we can keep tabs on it here in the repo.

This simply add a CircleCI config to trigger the osrf/ros2 night tags every night,
and the devel tag every day, as to spread out the build queue on docker hub.
It also triggers to build both if any changes are made to the repo.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,8 @@ commands:
             curl --request POST \
               --header "Content-Type: application/json" \
               --data << parameters.data >> \
-              --url << parameters.url >>
+              --url << parameters.url >> \
+            > /dev/null
           when: always
 
 executors:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ commands:
               --header "Content-Type: application/json" \
               --data << parameters.data >> \
               --url << parameters.url >> \
-            > /dev/null
+            > /dev/null && true
           when: always
 
 executors:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,12 +22,12 @@ references:
     trigger_docker_hub:
       url: $DOCKER_TRIGGER_URL_ROS2
       data: >-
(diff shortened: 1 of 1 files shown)


tfoote:
One question is that if we already have travis builds integrated. Why not use their cron functionality instead of relying on/integrating another provider?

https://docs.travis-ci.com/user/cron-jobs/

ruffsl:
> One question is that if we already have travis builds integrated. Why not use their cron functionality instead of relying on/integrating another provider?

I don't think travis has a way to create a cron job for a individual workflow. It's sort of all or nothing, and as our current docker build PR bot is using that as a trigger to check all tracked dockerfile, I didn't want to disrupt or refactor it to be cron job aware. Did travis add features for workflow selection?

tfoote:
If there's a reason that's fine. I just want to make sure that we're consciously choosing to add another integration.

ruffsl:
@nuclearsandwich or @tfoote , could you enable circleci for this repo now?

Additionally, sence DockerHub as deprecated the ability to trigger builds from linked repos that are official images, we may need to use this same cron to keep the desktop tags on osrf/ros update as well.

mikaelarguedas:
FYI: I triggered the nightly images manually (they didnt get triggered for 20 days). And opened an alternative to this at #348 to get a daily triggered build of those and email notification on failure.

mikaelarguedas:
Implemented in https://github.com/osrf/docker_images/pull/348

mikaelarguedas:
> Additionally, sence DockerHub as deprecated the ability to trigger builds from linked repos that are official images, we may need to use this same cron to keep the desktop tags on osrf/ros update as well.

This has been detailed in a new issue https://github.com/osrf/docker_images/issues/362#issue-558106322

Issue #313, opened 2019-08-28
Ship default colcon mixin and metadata with official ROS2 images and devel/nightly


ruffsl:
I think it would be nice if colcon mixin came pre-installed with default mixins already included for the official ROS2 images and for the devel and nightly images. How far back was colcon mixin first supported? ROS2 dashing? We could amend the template to install the tools for those releases.
cc @dirk-thomas @mikaelarguedas

mikaelarguedas:
This is shipped on all platforms targeted by ROS 2 named releases https://github.com/colcon/colcon-mixin/blob/f438905e74f9d57f381ffc66576cd567a344871b/stdeb.cfg#L4

> We could add it to all ros2 Dockerfiles

Originally the reason for not adding them to the docker images was to match the setup of the official installation instructions. So if we stick to that logic we would not add it to the ros:{ROS2_DISTRO} or `ros2:source` images.
Adding mixins to devel and nightly would be useful though.
Related is colcon-metadata that comes installed with colcon-common-extensions, we could set them up as well, e.g. to leverage https://github.com/colcon/colcon-metadata-repository/blob/4301f7737eec0a54723121c83f7409b6275a594c/fastrtps.meta#L4

I currently set  them up on downstream images: https://github.com/mikaelarguedas/ros2ci/commit/16372df9096e2cf6348a349a53473e4236ea3998

mikaelarguedas:
>  I'd recommend we update the instructions to include mixin.

That would work, we'd need to see how the ROS 2 team feels about it as it is not used/required by the instructions.

One concern I had to add it to images not updated regularly is that any change to the mixins / metadata need to be actively pulled.
For devel/nightly images it doesn't matter as they are rebuilt daily.
For the official images it is trickier as if a mixin/meta changes, the changes will not propagate resulting in possibly diverging behavior between a build based on nightly or an official image. This could fail in ways that would be hard for end users to track

mikaelarguedas:
That will be the best way for users to use it :+1:. Documenting that users should invoke the update commands for updating mixins and metadata in all their dockerfiles based on these images would be valuable.

While mixin should be only updated by users using them. metadata should be updated by all users as it is used by default (though metadatas have been changing very little since they were introduced, so the necessity to update should be low)

mikaelarguedas:
@ruffsl looks like this has been merged and rolled out.
Can this be closed or is there still some pending action items?

PR #315, opened 2019-08-28
Include colcon mixin and metadata in ros2 images

ruffsl:
Context: https://github.com/osrf/docker_images/issues/313


ruffsl:
@dirk-thomas, carrying over from https://github.com/osrf/docker_images/issues/313 do you have any opinion on this?
Should we go ahead and include colcon mixin and metadata with the official images?

dirk-thomas:
Sounds fine with me.

ruffsl:
relevant CI passing

PR #316, opened 2019-08-31
Add rosdep update options for EOLs

chikuta:
rosdep update command require additional option "--include-eol-distros" for eol packages like trusy.
This option was merged by this pull request (https://github.com/ros-infrastructure/rosdep/commit/ae0e89aa7feea3e73c9c3c94ff0f08522e6782b8) .

Please feel free to comment what to improve, I may have done something wrong as this is my first pull request.

### DIFF
--- a/ros/indigo/ubuntu/trusty/ros-core/Dockerfile
+++ b/ros/indigo/ubuntu/trusty/ros-core/Dockerfile
@@ -27,7 +27,7 @@ ENV LC_ALL C.UTF-8
 
 # bootstrap rosdep
 RUN rosdep init \
-    && rosdep update
+    && rosdep update --include-eol-distros
 
 # install ros packages
 ENV ROS_DISTRO indigo
--- a/ros/jade/ubuntu/trusty/ros-core/Dockerfile
+++ b/ros/jade/ubuntu/trusty/ros-core/Dockerfile
@@ -27,7 +27,7 @@ ENV LC_ALL C.UTF-8
 
 # bootstrap rosdep
 RUN rosdep init \
-    && rosdep update
+    && rosdep update --include-eol-distros
 
 # install ros packages
 ENV ROS_DISTRO jade


ruffsl:
We'll want to make sure the dockerfile changes propagate to images on docker hub as well.

mikaelarguedas:
> We'll want to make sure the dockerfile changes propagate to images on docker hub as well

This is not obvious to me, can you clarify why ?

From my perspective it looks like:
- the images on dockerhub had their rosdep database updated before going EOL. I would actually prefer NOT updating it as it reflects the state of the distro before it went down.
- not sure we can convince the librarian to reenable trusty and older builds for us just for this.

---
Related to the first bullet, breaking changes have been integrated in the rosdep database since the distros went EOL  (e.g. https://github.com/ros/rosdistro/pull/19843) and there is a discussion to how to safely update rosdep for EOL distros going on at https://github.com/ros/rosdistro/issues/18296

mikaelarguedas:
@ruffsl I opened https://github.com/osrf/docker_images/issues/327 as a discussion ticket for updating rosdep to EOL distros

ruffsl:
Ah, good points. I had forgot the old db is already archived within the image. If users want to invoke rosdep update downstream, then they'll need to tweek their Dockerfile to include the `--include-eol-distros` arg anyhow, but this doesn't impact the viability of the base images already archived. I was also thinking this only affects the legacy images, which are hosted entirely under the osrf docker hub org, but time has marched on such that EOL now include distros released in the official docker hub library.

> not sure we can convince the librarian to reenable trusty and older builds for us just for this.

No, probably not, nor do I think this would warrant us to necromance those official library builds.

mikaelarguedas:
Merging this to credit original contributor,
opened #327 and #330 as follow up

chikuta:
@mikaelarguedas @ruffsl 
My bad, I didn't notice these posts.
Thank you for your kindness. Next time, I will make a ticket at first.

PR #325, opened 2019-10-09
Add arm32 docker builds for dashing

AAlon:
Would be useful to generate arm32 docker containers like we do for ROS1 distros. 

Note - I really don't know if this is the only change required to enable arm32 dashing docker builds. If (significant) additional work is required I'll just open an issue instead and plan for it properly.

### DIFF
--- a/ros/manifest.yaml
+++ b/ros/manifest.yaml
@@ -403,6 +403,7 @@ release_names:
                         archs:
                             - amd64
                             - arm64v8
+                            - arm32v7
                         tag_names:
                             ros-core:
                                 aliases:


ruffsl:
> Would be useful to generate arm32 docker containers like we do for ROS1 distros.

This is conditional on the release of arm32v7 debian binaries for ROS2 Dashing. Last I check, Dashing doesn't have a tier support for the arm32v7 architecture. However, you could add your petition here:

https://discourse.ros.org/t/official-docker-images-for-ros2-crystal-and-arm32v7/6872

> Note - I really don't know if this is the only change required to enable arm32 dashing docker builds.

You'll have to change both the yaml and manifest file for ROS:

https://github.com/osrf/docker_images/blob/e8f80e9bdc35dff2cfe4a114d72fece8d23fb286/ros/ros#L105-L110

mikaelarguedas:
While preparing https://github.com/osrf/docker_images/pull/329 I noticed that there are now armhf debs for both Dashing and Eloquent.

So it should be possible to provide docker images for them.

@nuclearsandwich @tfoote Are you in favor of providing official docker images for non tier 1 platforms if they provide debs?

nuclearsandwich:
> @nuclearsandwich @tfoote Are you in favor of providing official docker images for non tier 1 platforms if they provide debs?

I'm not in favor of adding infrastructure on top of Dashing's armhf debs. armhf is not a tier 1 platform for Dashing and thus I can't promise that each sync will be free of known regressions. The story is a little different on Eloquent where armhf is still tier 2 but available from the outset. But whatever stability is expected of these docker images I would downgrade it for armhf.

ruffsl:
Finishing over at https://github.com/osrf/docker_images/pull/337

PR #328, opened 2019-10-20
Revert "rever ROS_PACKAGE_PATH hack allowing rosdep to find the ros2 packages"

mikaelarguedas:
Reverts osrf/docker_images#326

Reintroduces the RPP hack because the fix for https://github.com/ros-infrastructure/rosdep/issues/660 is incomplete leading consumers of the image to break.

Current rosdep behavior:
Packages in the ament_prefix_path that do install a package.xml but dont register themselves in the ament index fail to be detected


mikaelarguedas:
Ticketed question upstream at https://github.com/ros-infrastructure/rosdep/issues/724

PR #329, opened 2019-10-20
Add ROS2 Eloquent

mikaelarguedas:
Opening this for visibility pending release and armhf discussion reaching conclusion

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install:
 
 env:
   # ros2
+  - HUB_REPO=ros   HUB_RELEASE=eloquent HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=bionic
   - HUB_REPO=ros   HUB_RELEASE=dashing  HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=bionic
   - HUB_REPO=ros   HUB_RELEASE=crystal  HUB_OS_NAME=ubuntu HUB_OS_CODE_NAME=bionic
   # ros
--- a/ros/eloquent/ubuntu/bionic/Makefile
+++ b/ros/eloquent/ubuntu/bionic/Makefile
@@ -1,0 +1,28 @@
+all: help
+
+help:
+	@echo ""
+	@echo "-- Help Menu"
+	@echo ""
+	@echo "   1. make build            - build all images"
+	@echo "   2. make pull             - pull all images"
+	@echo "   3. make clean            - remove all images"
+	@echo ""
+
+build:
+	@docker build --tag=ros:eloquent-ros-core-bionic			ros-core/.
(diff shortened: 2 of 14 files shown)


mikaelarguedas:
Thanks Ruffin for updating this.

I see the official installation instructions are not allowing the installation of armhf Deb's by default.
Based on discussion from https://github.com/osrf/docker_images/pull/325, we need to find a way to communicate that the armhf images are not supported the same way the tier 1 architectures are.
I'd also like to emphasize some things in the eloquent images announcement because it appeared at ROSCon that users don't know we'll about them especially the ros1-bridge ones

ruffsl:
> I see the official installation instructions are not allowing the installation of armhf Deb's by default.

"not allowing"?
I've tested the arm32v7 images build locally, and am in favor of merging them upstream to make them more accessible for further testing, with the caveat that I'll update the official library docs to point out the tier platform rep.

> I'd also like to emphasize some things in the eloquent images announcement because it appeared at ROSCon that users don't know we'll about them especially the ros1-bridge ones

Sounds like another thing we could point out on https://github.com/docker-library/docs/pull/1381
Have a minimal compose example using the bridge you'd suggest there?

mikaelarguedas:
> "not allowing"?

The installation instructions explicitly list the architectures and the architecture list doesn't include armhf: 
```
sudo sh -c 'echo "deb [arch=amd64,arm64] http://packages.ros.org/ros2/ubuntu `lsb_release -cs` main" > /etc/apt/sources.list.d/ros2-latest.list'
```

> in favor of merging them upstream to make them more accessible for further testing

Me too

> I'll update the official library docs to point out the tier platform rep.

:+1: does any other place come to mind to emphasize the fact that it's tier 2? (I'd also highlight it in the announcement)

> Have a minimal compose example using the bridge you'd suggest there?

I don't but it should take just a couple minutes to create one.
I'll provide this as well as a review of the overall PR in a day or two

I was also hoping to have that as part of desktop by the time eloquent was released but this seems stale so I wouldn't hold the first batch of images for that.

ruffsl:
> +1 does any other place come to mind to emphasize the fact that it's tier 2? (I'd also highlight it in the announcement)

We could piggyback on the official eloquent announcement over discourse when this is upstreamed and pushed to registry. Looking forward to comments on the docs PR, it really needs to be merged.

ruffsl:
PR opened upstream: https://github.com/docker-library/official-images/pull/7039

mikaelarguedas:
Added the `desktop` and `ros1-bridge` tags to dockerhub and triggered a build

mikaelarguedas:
@ruffsl ros1-bridge is broken in Eloquent, so we should likely wait for the fix to be merged and synced before announcing these images

PR #345, opened 2019-12-20
remove COLCON_IGNORE at root of AMENT_PREFIX_PATH

mikaelarguedas:
This way if a consumer points the `ROS_PACKAGE_PATH` to `/opt/ros/$ROS_DISTRO` catkin_pkg can find packages.

pros: 
- brings us closer to the official docker images that don't have colcon files in their install space
- supports applications like industrial_ci that currently set the ROS_PACKAGE_PATH themselves

con:
- (more feature than a bug) crawling can happen if the `ROS_PACKAGE_PATH` is overriden, just like anywhere else on the system

Relates to https://github.com/ros-industrial/industrial_ci/pull/454#discussion_r360396351

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -62,7 +62,8 @@ ENV ROS_DISTRO foxy
 RUN mkdir -p /opt/ros/$ROS_DISTRO
 ARG ROS2_BINARY_URL=https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
 RUN wget -q $ROS2_BINARY_URL -O - | \
-    tar -xj --strip-components=1 -C /opt/ros/$ROS_DISTRO
+    tar -xj --strip-components=1 -C /opt/ros/$ROS_DISTRO \
+    && rm -f /opt/ros/$ROS_DISTRO/COLCON_IGNORE
 
 # Overwrite setup scripts with ones that point to /opt/ros/$ROS_DISTRO
 RUN mkdir -p /tmp/dir/build \


ruffsl:
Why does the nightly archive include the `COLCON_IGNORE` to begin with? Did it have ever have a purpose? Perhaps its worth fixing the archive upstream packaging  to remove it instead?

mikaelarguedas:
> Why does the nightly archive include the COLCON_IGNORE to begin with? Did it have ever have a purpose? 

It appears that colcon doesn't assume that the code is located in a `src` directory. So when it builds a workspace it drops a `COLCON_IGNORE` in any directory it creates. The goal is for the packages to NOT be detected there by colcon and avoid having errors like "there are 2 packages with same name in this workspace".
 
e.g. in the devel image that has an empty workspace:
```
$ docker run -it --rm osrf/ros2:devel
root@869d3ca7d73e:/opt/ros2_ws# ls
build  install  log  src
root@869d3ca7d73e:/opt/ros2_ws# find -name COLCON_IGNORE
./build/COLCON_IGNORE
./log/COLCON_IGNORE
./install/COLCON_IGNORE
```

In the case of the nightly archive I am not sure in which scenario the presence of the COLCON_IGNORE is useful.
I originally opened this here as a quick fix to unlock the use of `ros2:nightly` on industrial_ci. Now that it has been worked around, we can reconsider this PR.

Either:
- say "it doesnt matter because noone should be using catkin_pkg to crawl `/opt/ros/ROS_DISTRO` and rely on the ament resource index instead"
- it shouldn't be there to match the debs that are build without `colcon`:
  - ask upstream for the OK to remove it from the archive generation
    - if yes, discard this PR
    - if no, merge it

ruffsl:
> ask upstream for the OK to remove it from the archive generation
if yes, discard this PR
if no, merge it

That sounds good to me. Where do we open tickets for the "archive generation"?

mikaelarguedas:
Ticket open at https://github.com/ros2/ci/pull/362

mikaelarguedas:
https://github.com/ros2/ci/pull/362 landed

PR #348, opened 2019-12-22
CRON job to trigger daily build of ros2 images

mikaelarguedas:
ros2 images have not been rebuilt since Dec 4th (20 days ago).
It was suggested in the past to have a cron job to trigger these builds daily to not rely on the currently flaky setup. The PR for it has been blocked for a while https://github.com/osrf/docker_images/pull/312

This PR implements a cron job using Github Actions thus allowing us to use it without extra permissions given to a third party app.

How it compares to https://github.com/osrf/docker_images/pull/312:

- the secrets have to be defined in this repo. They are encrypted by github and filtered out of the logs
  - cons:
    - any matching string in the logs will be redacted, so it could actually leak password if the same string is printed to the logs from code (though unlikely). I believe the logs are not public, so leaking is prevented
    - Logs are not public?
- the images are built in the actions and not on dockerhub:
  - pro:
    - we get notified on build failure
    - the job starts instantly
  - cons:
    - slower (maybe?) as pulling, building and pushing will require some bandwidth and latency is likely lower on dockerhub
    - I'm discovering github actions and hopefully things can be optimized and improved later on

Leaving this in draft for a couple days the time it proves to run reliably at https://github.com/mikaelarguedas/docker_images/actions

### DIFF
--- a/.github/workflows/cron.yaml
+++ b/.github/workflows/cron.yaml
@@ -15,7 +15,7 @@ jobs:
     - uses: actions/checkout@v1
     - name: Call devel webhook
       run: |
-        curl -H "Content-Type: application/json" --data '{"docker_tag": "${{env.dockerTag}}"}'  -X POST ${{ secrets.DOCKER_POST_PUSH_TRIGGER }}
+        curl -H "Content-Type: application/json" --data '{"docker_tag": "${{env.dockerTag}}"}'  -X POST ${{ secrets.DOCKER_ROS2_POST_PUSH_TRIGGER }}
   nightly_build:
     name: Trigger nightly image
     env:
@@ -25,4 +25,4 @@ jobs:
     - uses: actions/checkout@v1
     - name: Call nightly webhook
       run: |
-        curl -H "Content-Type: application/json" --data '{"docker_tag": "${{env.dockerTag}}"}'  -X POST ${{ secrets.DOCKER_POST_PUSH_TRIGGER }}
+        curl -H "Content-Type: application/json" --data '{"docker_tag": "${{env.dockerTag}}"}'  -X POST ${{ secrets.DOCKER_ROS2_POST_PUSH_TRIGGER }}
--- a/.github/workflows/cron.yaml
+++ b/.github/workflows/cron.yaml
@@ -1,0 +1,28 @@
+---
+name: Trigger nightly images
+
+on:
+  schedule:
+    # nightly packaging job usually finish around 14 UTC
+    - cron:  '0 15 * * *'
+jobs:
+  source_build:
+    name: Trigger devel image
+    env:
+      dockerTag: ${{ 'devel' }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Call devel webhook
+      run: |
+        curl -H "Content-Type: application/json" --data '{"docker_tag": "${{env.dockerTag}}"}'  -X POST ${{ secrets.DOCKER_POST_PUSH_TRIGGER }}
+  nightly_build:
+    name: Trigger nightly image
+    env:
+      dockerTag: ${{ 'nightly' }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Call nightly webhook
+      run: |
+        curl -H "Content-Type: application/json" --data '{"docker_tag": "${{env.dockerTag}}"}'  -X POST ${{ secrets.DOCKER_POST_PUSH_TRIGGER }}
--- a/.github/workflows/trigger_devel.yaml
+++ b/.github/workflows/trigger_devel.yaml
@@ -1,0 +1,17 @@
+---
+name: CRON Trigger devel images
+
+on:
+  schedule:
+    - cron:  '0 3 * * *'
+jobs:
+  source_build:
+    name: Trigger devel image
+    env:
+      dockerTag: ${{ 'devel' }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Call devel webhook
+      run: |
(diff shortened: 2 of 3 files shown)


ruffsl:
I see you've opted to have the docker engine in the CI build in image, instead of that of the automated registry. One reason in #312 that I opted to use the registry API to trigger the build was that it would subsequent trigger any automated repo build pointed at `osrf/ros2` repo upon success. Could you verify that pushing remotely to the registry instead still triggers downstream docker hub repos to rebuild?

mikaelarguedas:
> was that it would subsequent trigger any automated repo build pointed at osrf/ros2 repo upon success

Oh right, it's definitely a potential downside, I havent tried it but I would expect dockerhub to trigger downstream repo only from successful automated builds (and not from image pushes).
I'll give it a shot soon and report here.

If we end up having to back to triggering them instead of pushing them, just saw that triggering `nightly` yeasterday did not trigger the downstream `nightly-rmw*` images .. So we'll need to trigger all images from the cron job.

ruffsl:
> just saw that triggering nightly yeasterday did not trigger the downstream nightly-rmw* images

Hmm, that was not what I've seen in the past, as this post_push hook was working when I setup the the docker hub repo settings. You triggered the `nightly` tag, yet on build/push completion by docker hub, the webhook did not trigger/cascade to build the next tag?
 
https://github.com/osrf/docker_images/blob/c857f6c9bae5fd71e3f71739e6c9727bf5f497cf/ros2/nightly/nightly/hooks/post_push#L1-L4

ruffsl:
Looks like the trigger api URL has migrated from `https://cloud.docker.com/api/build/v1/source/.../call` to `https://hub.docker.com/api/build/v1/source/.../call`. 

I've updated the URL in the `POST_PUSH_TRIGGER` for the `BUILD ENVIRONMENT VARIABLES`.

ruffsl:
Also, looks like the github link to the `osrf/docker_images` repo has been unset. It doesn't seem like I have the github permissions to re-link it: ping @tfoote or @nuclearsandwich 

image


This perhaps explains why my fork with the CircleCI cron job (using https://github.com/osrf/docker_images/pull/312) triggering was ineffective while still correctly calling the API daily.

https://circleci.com/gh/ruffsl/docker_images/tree/master

mikaelarguedas:
> Also, looks like the github link to the osrf/docker_images repo has been unset.

This seem to be fixed
Screenshot from 2019-12-24 09-15-15

mikaelarguedas:
> was that it would subsequent trigger any automated repo build pointed at osrf/ros2 repo upon success

It actually works, downstream repos get triggered when new mages are pushed :tada:. tested on [this repo](https://hub.docker.com/repository/docker/mikaelarguedas/test_docker_downstream_trigger/builds)

(unrelated to the push/build on dockerhub)Note that it works only with repos where the base image is explicitly set: `FROM osrf/ros2:devel`. Doesn't work for files that have a configurable base image 
```
ARG FROM_IMAGE=osrf/ros2:devel
FROM $FROM_IMAGE
```

One possible downside of this approach if that the build logs will not be on docker hub anymore. Not a big bummer though

---

I also thought a bit about the triggering approach and a few questions come up compared to this github actions approach:
- How would we confirm that the build trigger worked? (to avoid surprises like https://github.com/osrf/docker_images/pull/348#issuecomment-568603804)
- How would we get notified if the automated build fails? (there's no webhook for build failure on DockerHub).
  - Potential solution: We can enable email notifications on build failure. These are per organization and not per repo, so we'll receive email for any failed builds on the `osrf` org(currently the only one failing is [sros](https://hub.docker.com/repository/docker/osrf/sros/builds). It may be worth enabling regardless of the approach we take.

mikaelarguedas:
> Potential solution: We can enable email notifications on build failure. These are per organization and not per repo, so we'll receive email for any failed builds on the osrf org (currently the only one failing is sros). It may be worth enabling regardless of the approach we take.

https://github.com/osrf/docker_images/pull/350 fixes the build of the sros image

mikaelarguedas:
Follow-up of offline discussion:

- Email notifications on dockerhub have been enabled after thumbsup from Tully
- Made a change to go back to triggering builds on dockerhub insteal of building in github actions https://github.com/mikaelarguedas/docker_images/commit/88e3569d53c1e632a606932da32e1f5d003d02ad

I'll let it run for a day or 2 before forcepushin the change to this branch

ruffsl:
> I'll let it run for a day or 2 before forcepushin the change to this branch

Ok, I've just now disabled the CircleCI on my fork to check yours is working fine.

mikaelarguedas:
> Ok, I've just now disabled the CircleCI on my fork to check yours is working fine.

:+1: Seems to have ran successfully today.

Regarding the implementation. I couldn't get curl to expand `$dockerTag` if it was provided as an env var in the build matrix. So I had to duplicate the jobs... Now this workflow has 2 jobs one to trigger the devel image and one for the nightly one

.github/workflows/cron.yaml
```diff
@@ -0,0 +1,28 @@
+---
+name: Trigger nightly images
+
+on:
+  schedule:
+    # nightly packaging job usually finish around 14 UTC
+    - cron:  '0 15 * * *'
```
ruffsl: One thing the CircleCI version I wrote did was to trigger devel and nightly tags at different times of the day, as to reduce the queuing time for the entirety of the nightly tags, so that they all finished together as ASAP.

mikaelarguedas: From my understanding of GH Action triggers, to do this we'll need 2 different workflows/yaml files. This is not hard to do though, and may be worth doing as the devel image can be built at any time of the day (see https://github.com/osrf/docker_images/pull/348/commits/c788ed32fe6dd7caf3e56b1aa4133201558adc04).

We'll still have some time between the various nightly tags though, and this can lead to inconsistencies for people running jobs using several of them during that period...

I guess that was an unexpected advantage of building them in the ci and pushing them.

PR #349, opened 2019-12-24
remove ros_workspace, use COLCON_CURRENT_PREFIX

mikaelarguedas:
Sourcing of ROS 1 underlay has been removed from the setup files in the archives: https://github.com/ros2/ci/pull/356

So the setup files can be sourced by simply defining `COLCON_CURRENT_PREFIX` in the image.

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -64,6 +64,9 @@ RUN mkdir -p /opt/ros/$ROS_DISTRO
 ARG ROS2_BINARY_URL=https://ci.ros2.org/view/packaging/job/packaging_linux/lastSuccessfulBuild/artifact/ws/ros2-package-linux-x86_64.tar.bz2
 RUN wget -q $ROS2_BINARY_URL -O - | \
     tar -xj --strip-components=1 -C /opt/ros/$ROS_DISTRO
+
+# Overwriting _colcon_prefix_chain_sh_COLCON_CURRENT_PREFIX to point to the new install location
+# Necessary because the default value is an absolute path valid only on the build machine
 RUN sed -i "s|^\(_colcon_prefix_chain_sh_COLCON_CURRENT_PREFIX\s*=\s*\).*$|\1/opt/ros/$ROS_DISTRO|" \
       /opt/ros/$ROS_DISTRO/setup.sh
 
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -39,7 +39,6 @@ ENV ROS_DISTRO foxy
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 ENV ROSDISTRO_INDEX_URL https://raw.githubusercontent.com/osrf/docker_images/master/ros2/nightly/nightly/index-v4.yaml
-ENV COLCON_CURRENT_PREFIX /opt/ros/$ROS_DISTRO
 
 # install python packages
 RUN pip3 install -U \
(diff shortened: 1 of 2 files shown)


ruffsl:
I'll test this again before merging

ruffsl:
I tested building navigation2 against this but looks like it revealed an issue with rosdep missing cmake packages in the underlay like we talked about offline:

https://github.com/ros-planning/navigation2/blob/4cf60e9d342adf71bef5c0a60faa7c686c0f86fa/Dockerfile

```
Step 27/31 : RUN . $UNDERLAY_WS/install/setup.sh &&     apt-get update && rosdep install -q -y       --from-paths src       --ignore-src     && rm -rf /var/lib/apt/lists/*
 ---> Running in 00bea3a2a3da
[connext_cmake_module] Warning: The location at which Connext was found when the workspace was built [[/opt/rti.com/rti_connext_dds-5.3.1]] does not point to a valid directory, and the NDDSHOME environment variable has not been set. Support for Connext will not be available.
[connext_cmake_module] Warning: The location at which Connext was found when the workspace was built [[/opt/rti.com/rti_connext_dds-5.3.1]] does not point to a valid directory, and the NDDSHOME environment variable has not been set. Support for Connext will not be available.
Get:1 http://packages.ros.org/ros2-testing/ubuntu bionic InRelease [3947 B]
...
Get:20 http://archive.ubuntu.com/ubuntu bionic-backports/universe amd64 Packages [4241 B]
Fetched 18.1 MB in 4s (5012 kB/s)
Reading package lists...
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
nav2_behavior_tree: Cannot locate rosdep definition for [behaviortree_cpp_v3]
dwb_critics: Cannot locate rosdep definition for [angles]
nav2_system_tests: Cannot locate rosdep definition for [gazebo_ros_pkgs]
dwb_plugins: Cannot locate rosdep definition for [angles]
nav2_bt_navigator: Cannot locate rosdep definition for [behaviortree_cpp_v3]
```

mikaelarguedas:
oh right actually this fix will not work when there are multiple levels of overlays.

I guess we could go back to sed'ing the setup.sh. This time it should be more minimal as there is only one variable to replace.
added https://github.com/osrf/docker_images/pull/349/commits/1507da4cbcae2fc7ed8577dfd550d4bd8c225978 for an example of what it would look like

ros2/nightly/nightly/Dockerfile
```diff
@@ -59,21 +60,12 @@ RUN pip3 freeze | grep pytest \
...
-# Overwrite setup scripts with ones that point to /opt/ros/$ROS_DISTRO
-RUN mkdir -p /tmp/dir/build \
- && cd /tmp/dir \
- && git clone --depth 1 https://github.com/ros2/ros_workspace.git -b latest \
- && cd /tmp/dir/build \
- && COLCON_CURRENT_PREFIX=/opt/ros/$ROS_DISTRO . /opt/ros/$ROS_DISTRO/local_setup.sh \
- && cmake -DCMAKE_INSTALL_PREFIX=/opt/ros/$ROS_DISTRO ../ros_workspace \
- && make install \
- && rm -r /tmp/dir
+RUN sed -i "s|^\(_colcon_prefix_chain_sh_COLCON_CURRENT_PREFIX\s*=\s*\).*$|\1/opt/ros/$ROS_DISTRO|" \
```
ruffsl: Could you add back line break and a comment to remind the reader of purpose for this sed? Perhaps a link to an old rel relevant issue, or just a sentence.

PR #350, opened 2019-12-26
[sros] rename roslisp manifest.xml for rosdep

mikaelarguedas:
Follow-up of https://github.com/osrf/docker_images/commit/cb860f76b6146049aff4b080f40b316a70501f56 to get sros images building again.

This is a workaround because of a weird behavior of rosdep that;s ticketed at https://github.com/ros-infrastructure/rosdep/issues/736

With this merge all images in this repo build successfully, and we should be able to enable email notifications on dockerhub :tada:

### DIFF
--- a/sros/kinetic/Dockerfile
+++ b/sros/kinetic/Dockerfile
@@ -53,8 +53,11 @@ RUN rosinstall_generator \
     git clone -b sros https://github.com/ros/ros_comm && \
     git clone -b sros https://github.com/ros-infrastructure/rospkg ../rospkg
 
+# FIXME renaming manifest.xml is a workaround for rosdep misbehaving when both
+# manifest.xml and package.xml are present
 # install dependencies
-RUN apt-get update && \
+RUN mv roslisp/manifest.xml roslisp/oldmanifest.xml && \
+    apt-get update && \
     rosdep init && \
     rosdep update && \
     rosdep install -y \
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Why not remove the `manifest.xml` file in the upstream repo instead? The latest ROS version using `manifest.xml` files has long been EOLed already.

ruffsl:
I been thinking of cleaning/removing the sros1 tag for a while. Perhaps we should just retire this folder.

mikaelarguedas:
> Why not remove the manifest.xml file in the upstream repo instead? The latest ROS version using manifest.xml files has long been EOLed already.

Mostly because the changes required in the upstream repo to get rid of rosbuild remnant may be more involved than removing a single file. And getting it released all the way back to kinetic seems unlikely or at least not quickly.

So I sided with reporting the issue on rosdep as this seems like a case where rosdep should behave differently, and patching here with a nonintrusive patch just to get the things building again.

> I been thinking of cleaning/removing the sros1 tag for a while. Perhaps we should just retire this folder.

@ruffsl Should we disable the builds on osrf/sros then ?

ruffsl:
> Should we disable the builds on osrf/sros then ?

I've just now disabled the automated build for osrf/sros.

mikaelarguedas:
> I've just now disabled the automated build for osrf/sros.

Do you prefer to merging this so that the file builds or closing this PR?

PR #355, opened 2020-01-17
Use kmod to fix gzclient images

mikaelarguedas:
This fixes the build on gzclient images on Debian.

We may want to revisit the entrypoint / command of these images because currently they just run `gzserver` as their parents. I'm tempted to just make these call `gazebo` and keep the entrypoint of the base image

### DIFF
--- a/gazebo/.config/images.yaml.em
+++ b/gazebo/.config/images.yaml.em
@@ -50,8 +50,8 @@ images:
             - docker_templates
         upstream_packages:
             - binutils
+            - kmod
             - mesa-utils
-            - module-init-tools
             - x-window-system
         gazebo_packages:
             - gazebo@(gazebo_version)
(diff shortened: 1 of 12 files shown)


ruffsl:
Why is kmod needed, isn't that for managing kernel modules?

mikaelarguedas:
> Why is kmod needed, isn't that for managing kernel modules?

it is. As of why it's needed I can't really speak of that. IDK why [`module-init-tools`](https://packages.ubuntu.com/bionic/module-init-tools) was listed as a dependency so I made the change that would keep the same behavior while fixing the debian builds.

mikaelarguedas:
Users can run gzclient using the official images. e.g.:

```
rocker --x11 --nvidia  gazebo:gzserver10 gazebo
```

So there is no benefit in building the gzclient images
Rejecting this PR

Issue #358, opened 2020-01-29
osrf/ros2 docker for ros2 doesn't work


claireyywang:
https://hub.docker.com/r/osrf/ros2/ this is the first search result on google when I searched for `docker ros2` and the command `docker pull osrf/ros2` does not work. 
```
Using default tag: latest
Error response from daemon: manifest for osrf/ros2:latest not found: manifest unknown: manifest unknown
```
This page is misleading. I believe the actual docker page for all ros related images is this one https://hub.docker.com/_/ros . There have been four similar issues on this matter, which I think proves the amount of traffic on the site. Can we either update the pull command, or add information on the page to redirect users to where they are supposed to go?

mikaelarguedas:
Closing as duplicate of https://github.com/osrf/docker_images/issues/186

mikaelarguedas:
Sorry for the premature closing I read the description too fast.

The rationale for not having a latest tag is highlighted on #186. 

> Can we either update the pull command,

Unfortunately I'm not aware of a way to change the command shown for pulling.

> or add information on the page to redirect users to where they are supposed to go?

The header of the webpage is "**Experimental**, Docker images for ROS 2 development only"
And below is a link to the ROS official images.
I added "No `latest` tag, pull with *explicit* tags" to the header but couldn't come up with something clearer given the description's 100-character limit.

Do you have a specific suggestion or wording in mind that could make it clearer for users landing on the page ?

claireyywang:
> The header of the webpage is "**Experimental**, Docker images for ROS 2 development only"
> And below is a link to the ROS official images.

To be honest the `ROS` link didn't even stand out to me as a link. I had to go back and look for it after reading your comment. And to someone not familiar with docker but just want to get ros2 up and running, it requires quite a bit of reading to tell which one of the two sites is supposed to work. 

> Do you have a specific suggestion or wording in mind that could make it clearer for users landing on the page ?

Maybe edit the header or the description (very top underneath `OSRF Docker Images`) to say something like 

" This repo contains images specific to ROS 2, and are particularly for development and continuous integration. These may contain experimental features or unstable interfaces for testing the latest ROS 2 branches. Go to  https://hub.docker.com/_/ros for official stable releases of ROS and ROS 2 images. "

Just to give people an idea of what this page has and how it is different from  https://hub.docker.com/_/ros page, and that would help them decide which one they want. 

[...]

Issue #186, opened 2018-08-22
docker pull access denied for osrf


kirk86:
Hi, thank you for publishing theses images.
I am trying to pull ros2 using docker `docker pull osrf/ros2` but I keep getting the following error message
```
Using default tag: latest
Error response from daemon: manifest for osrf/ros2:latest not found
```
Am I doing something wrong here. How are these images supposed to be used? Is there any other tag other than `latest` that we should use?

Thank you in advance!

tfoote:
That's not a access denied, but a failure to resolve. We don't have a `latest` tag in the repository which is the default alias. 

You should pick from one of the tags listed here: https://hub.docker.com/r/osrf/ros2/tags/ such as `bouncy-full`

I'll leave this open as we should probably pick one to be the default and duplicate it as latest so others don't hit this issue.

kirk86:
@tfoote Thank you so much that clarifies things a lot. If you don't mind me asking a second question when we pull ros2 and get a tty to it does gazebo come preinstalled in that ros2 version or do we have to do separate pull for that?

ruffsl:
> we should probably pick one to be the default and duplicate it as latest so others don't hit this issue.

I'd recommend not setting that expectation, given that we don't do such for the `osrf/ros` (ros1) images. I'd imagine after ros2 crystal is released, we can migrate crystal (and possibly past releases) into the official docker library, and leave just the desktop tags hosted on the `osrf` org. Otherwise, I think the abeguiry in there being both a `osrf/ros2:latest` and a `ros2:latest` might be confusing.

> does gazebo come preinstalled in that ros2 version

@kirk86 , gazebo is installed in ros1 image of `osrf/ros:<distro>-desktop-full` given that meta package includes simulation tools, but no such equivalent exists for ros2. You can however write your own custom dockerfile the builds from the ros2 image and installed gazebo on top, although I'm not sure a ro2 gazebo interface exists unless you'd want to leverage ros1-bridge.

mikaelarguedas:
> I'd recommend not setting that expectation, given that we don't do such for the osrf/ros (ros1) images. I'd imagine after ros2 crystal is released, we can migrate crystal (and possibly past releases) into the official docker library, and leave just the desktop tags hosted on the osrf org. Otherwise, I think the abeguiry in there being both a osrf/ros2:latest and a ros2:latest might be confusing.

:+1: that was my thinking as well. We will create a `latest` tag once we migrate the ros2 images to the docker library.

[...]

PR #363, opened 2020-01-31
Add table of contents to readme

mikaelarguedas:
In an attempt to make the page more user-friendly for people reading it on dockerhub
Related to #358 and #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you require extra dependencies not available from the Official Library images
 
 ### [ROS ![Docker Pulls](https://img.shields.io/docker/pulls/osrf/ros.svg?label=pulls) ![Docker Stars](https://img.shields.io/docker/stars/osrf/ros.svg?label=stars)](https://hub.docker.com/r/osrf/ros/)
 
-This repo builds `FROM` the Official Library by including additional meta-packages such for desktop installations.
+This repo builds ROS (both ROS 1 and ROS 2) images `FROM` the Official Library by including additional meta-packages such for desktop installations.
 These images may be helpful using containers for development or when display forwarding when using GUIs.<br/>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,15 @@
...
 This repo hosts Dockerfiles and scaffolding for images in the Official Library and OSRF Organization on Docker Hub.
 
+## Table of Contents
+[Official docker library](#official-library)<br/>
+[Official ROS (1 and 2) images](#ros--)<br/>
+[Official Gazebo images](#gazebo--)<br/>
+[OSRF ROS (1 and 2) overlay images](#ros---1)<br/>
+[OSRF ROS 2 development images](#ros2--)<br/>
+[OSRF Gazebo overlay images](#gazebo---1)<br/>
+[OSRF ROS Legacy images](#ros-legacy--)<br/>
```
mikaelarguedas: Note to self: these will need to be made absolute before merging otherwise I think it will not work on dockerhub

ruffsl:
Perhaps we could just point out/link to the tags page on the respective docker hub repo, rather than hardcoding what tags do or do not yet exist? I think educating docker users about the conventions of latest is out of scope for top level readme.

mikaelarguedas:
Yeah that was my original feeling as well.

As people seem to be confused by the lack of latest (and most popular dockerhub repos have a latest) I figured maybe we were enough outside of the "default" to justify explaining it after all.
But just mentionning the command format and point to the tag list may be enough :+1:

How about something like this:
"""
Images must be pulled using an explicit tag e.g:
```
docker pull osrf/ros:<tag_name>
```
List of tags available at https://hub.docker.com/r/osrf/ros/tags
"""

ruffsl:
That looks good.

PR #371, opened 2020-02-05
Add Gazebo11

mikaelarguedas:
Add gazebo 11 dockerfiles.

Currently this targets only Ubuntu Bionic and only amd64 architecture.

@chapulina @j-rivero Do you know if there other OSes and / or architectures that Gazebo11 is targeting / released for ?

### DIFF
--- a/.github/workflows/gazebo_ci.yaml
+++ b/.github/workflows/gazebo_ci.yaml
@@ -20,6 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         env:
+          - {HUB_REPO: gazebo, HUB_RELEASE: 11, HUB_OS_NAME: ubuntu, HUB_OS_CODE_NAME: bionic}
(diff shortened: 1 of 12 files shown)


j-rivero:
> @chapulina @j-rivero Do you know if there other OSes and / or architectures that Gazebo11 is targeting / released for ?

At least Ubuntu Focal (when released) for amd64/i386 and Debian Sid amd64/i386. I will try to get Debian Buster too but I'm unsure of how would be the compatibility level for required toolchain and dependencies requirements. Please note that arm builds are impacted [by a couple of compilation bugs](https://bitbucket.org/osrf/gazebo/issues/2691/compilation-problems-for-arm32-and-armhf).

mikaelarguedas:
Thanks Jose for the update.
I'll submit this to the official docker library now and we'll add new platforms as they come out :+1:

Issue #258, opened 2019-04-08
Gazebo v9 build fails for arm32v7


ruffsl:
It looks like the latest release for Gazebo v9 has not been synced for arm32v7, and is blocking docker librage image builds for that architecture:

https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/gazebo/21/console 

```
E: Version '9.8.0-1*' for 'gazebo9' was not found
```

Is there an upstream regression that is blocking this Gazebo release?

arm32v7 [![Build Status](https://doi-janky.infosiftr.net/buildStatus/icon?job=multiarch/arm32v7/gazebo)](https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/gazebo/)

ruffsl:
@j-rivero , any update? Looks like this is still unresolved.

mikaelarguedas:
Looks like there are now 5 cases with missing packages:
gazebo 9 arm64 (stretch + bionic)
gazebo 9 arm32 (stretch + bionic)
gazebo 10 arm32 (bionic)

Are new gazebo packages still being released for arm ?

mikaelarguedas:
@chapulina do you happen to have any insight on this ?

j-rivero:
> Are new gazebo packages still being released for arm ?

arm64 and arm32 on Ubuntu suffer from a couple of bugs, I've created the public issue for this: https://bitbucket.org/osrf/gazebo/issues/2691/compilation-problems-for-arm32-and-armhf . Patches are more than welcome :)

mikaelarguedas:
Sounds good thanks for the update. I don't think I'll have time to look into this but I'll watch the issue.

In the meantime I suggest to remove the arm images from dockerhub to avoid them trying to build them daily.

ruffsl:
I think we can reopen this later if there's ever enough interest in arm for gazebo.

PR #374, opened 2020-02-05
[gazebo] Remove arm archs

mikaelarguedas:
Based on discussion https://github.com/osrf/docker_images/issues/258

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,10 @@ For complete listing of tag, view the Repo Info link below.
 | Type | Status |
 |---|---|
 | [amd64](https://hub.docker.com/r/amd64/gazebo/tags) | [![Build Status](https://doi-janky.infosiftr.net/buildStatus/icon?job=multiarch/amd64/gazebo)](https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/gazebo/) |
+<!--
 | [arm32v7](https://hub.docker.com/r/arm32v7/gazebo/tags) | [![Build Status](https://doi-janky.infosiftr.net/buildStatus/icon?job=multiarch/arm32v7/gazebo)](https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/gazebo/) |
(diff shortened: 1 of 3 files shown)


ruffsl:
Will we be notified when the arm builds are updated? I just like to avoid forgetting to revert this, then never having the arm images updated even after the gazebo packages are sync'ed for arm. Though the ci job for that architecture is failing https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/gazebo/ , at least we don't have to do anything on our end, other than just ensure the registry hashes are up to date.

mikaelarguedas:
> I'd just like to avoid forgetting to revert this,

That is a good point
By watching https://bitbucket.org/osrf/gazebo/issues/2691/compilation-problems-for-arm32-and-armhf we will be notified when the status changes.

As these have been failing for a year with no expectation of change in the near future, I'd rather us actively enabling when it is actually supported than consuming hours of build time daily to provide no output or once in a while unsupported content.

In the end, if gazebo doesn't support their releases on arm, I believe we should not do it either.

j-rivero:
> In the end, if gazebo doesn't support their releases on arm, I believe we should not do it either.

I think both of your are right. Makes little sense to support arm docker images until we fix the different bugs affecting the build. To mitigate the problem raised by Ruffin I [dropped a comment ](https://bitbucket.org/osrf/gazebo/issues/2691/compilation-problems-for-arm32-and-armhf#comment-56063781)on the Gazebo issue so we don't forget to come back to this repo once fixed.

MaximeAubanel:
So as of now, there is no way to pull osrf/ros:eloquent-ros1-bridge for arm ?

mikaelarguedas:
> So as of now, there is no way to pull osrf/ros:eloquent-ros1-bridge for arm ?

This is correct, although unrelated to this PR (ros1_bridge and gazebo are independant images) so the fact that gazebo arm compilation problems get resolved won't impact the ros1-bridge images.

Dockerhub doesnt provide ways to compile images for arm on third party profiles (like the osrf one). Only the official docker library has support for that.
So it is not possible to pull *any* images from `osrf/*` for arm.

As ros1_bridge doesn't require any GUI tools, we could submit that image to the official docker library and get arm images built. I opened https://github.com/osrf/docker_images/issues/414 to discuss it

PR #377, opened 2020-02-20
Remove unnecessary argcomplete from pip

mikaelarguedas:
Version of argcomplete in ubuntu repository in bionic is 1.8.1 that is high enough for the use in colcon/ros2cli. No need to install it from pip anymore.

### DIFF
--- a/ros/.config/ros2/images.yaml.em
+++ b/ros/.config/ros2/images.yaml.em
@@ -7,8 +7,6 @@ images:
         maintainer_name: @(maintainer_name)
         template_name: docker_images_ros2/create_ros_core_image.Dockerfile.em
         entrypoint_name: docker_images_ros2/ros_entrypoint.sh
-        pip3_install:
-            - argcomplete
         template_packages:
             - docker_templates
         ros2_packages:
--- a/ros/dashing/ubuntu/bionic/images.yaml.em
+++ b/ros/dashing/ubuntu/bionic/images.yaml.em
@@ -7,8 +7,6 @@ images:
         maintainer_name: @(maintainer_name)
         template_name: docker_images_ros2/create_ros_core_image.Dockerfile.em
         entrypoint_name: docker_images_ros2/ros_entrypoint.sh
-        pip3_install:
-            - argcomplete
         template_packages:
             - docker_templates
         ros2_packages:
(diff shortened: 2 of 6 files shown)


ruffsl:
I there a way we can be sure that 1.8.1 is recent enough? For instance, I'll be sad about needing to side load with pip downstream, as fish support was only realised as of 1.10.0 https://github.com/kislyuk/argcomplete/issues/68#issuecomment-491613751 . I wouldn't stall this if it's just fish shell though.

mikaelarguedas:
> I there a way we can be sure that 1.8.1 is recent enough?

Unfortunately no. Neither colcon nor ROS packages declare a minimum dependency on argcomplete. The main reason is because they implement fallback behaviors if the feature they try to use is not available in the installed version. So there is no minimal version as the code will work regardless, only less features will be available.

`fish` not being a supported shell, I would assume it's acceptable for users of the docker images not using the default shell to install specific software for their shell.

The only feature that is currently used in ROS 2 that is not available in the Ubuntu package is the SuppressCompleter used by ros2 run:
https://github.com/kislyuk/argcomplete/pull/224
More context at: https://github.com/ros2/ros2cli/pull/41#issuecomment-322335703
This would impact only packages installing a single executables, so it would be a small number of packages but would still impact some e.g. rviz2
before the change:
```
ros2 run rviz2 <TAB>
ros2 run rviz2 rviz2
```
after the change:
```
ros2 run rviz2 <TAB>
ros2 run rviz2 
--prefix  rviz2
```

If we consider this a deal breaker we can hold this change and apply it only to foxy (that has argcomplete 1.11.0).

I personally think its acceptable behavior in the docker images. Especially as users dont get any argcomplete completion out of the box when entering the container, they need to source setup files by hand to enable it.

ruffsl:
> The only feature that is currently used in ROS 2 that is not available in the Ubuntu package is the SuppressCompleter used by ros2 run: kislyuk/argcomplete#224

That SuppressCompleter sounds useful for a related issue I had: https://github.com/ros2/launch_ros/issues/118

> I personally think its acceptable behavior in the docker images. Especially as users dont get any argcomplete completion out of the box when entering the container, they need to source setup files by hand to enable it.

👍

Issue #379, opened 2020-03-05
OpenSplice should be removed from Foxy images


dirk-thomas:
Since it is officially not supported in Foxy.

mikaelarguedas:
Thanks for the headsup

There are currently no Foxy images.
The Foxy images once released will simply install "ros-foxy-ros-core". So it will come only with whatever the default RMW implementation is.

The `osrf/ros2:nightly` images track the state of the nightly archive. So they came without `rmw_opensplice` and related packages since https://github.com/ros2/ros2/pull/865

As far as I can tell there is no action item associated with this, I'm going to close this.

@dirk-thomas feel free to re-open if I misunderstood your comment and there is an action item associated with this issue

dirk-thomas:
I though there are Docker images tracking the current nightly builds from master (which is essentially Foxy).

dirk-thomas:
E.g. this would be obsolete: https://github.com/osrf/docker_images/blob/58f2e8a1d7bcf64cbe7ff0de936d58d4130eb918/ros2/nightly/nightly/Dockerfile#L99

mikaelarguedas:
> I though there are Docker images tracking the current nightly builds from master (which is essentially Foxy).

Yes the `ros2:nightly` docker image tracks the nightly builds of ci.ros2.org. But those builds did not include any Opensplice related packages since more than a month AFAICT.
Does https://github.com/ros2/ci/pull/401 impact the content of the nightly archive?


> E.g. this would be obsolete: 

Yeah I went back and forth on this one. If people pass a specific [ROS2_BINARY_URL](https://github.com/osrf/docker_images/blob/58f2e8a1d7bcf64cbe7ff0de936d58d4130eb918/ros2/nightly/nightly/Dockerfile#L64) they might still need it. 
Similarly to the source image [still skipping the `libopensplice67`](https://github.com/osrf/docker_images/blob/58f2e8a1d7bcf64cbe7ff0de936d58d4130eb918/ros2/source/images.yaml.em#L50) even if the last distro used `libopensplice69`

But there is a bigger question with this is that now we have `ros2:nightly` and `ros2:nightly-rmw` that are the same because the archive now ship with both fully fledged FastRTPS and CycloneDDS.
I opened https://github.com/osrf/docker_images/issues/380 to track those down

dirk-thomas:
> Does ros2/ci#401 impact the content of the nightly archive?

Probably not since the OpenSplice repo had already been removed for over a month.

[...]

PR #381, opened 2020-03-05
switch devel and nightly images to focal

mikaelarguedas:
~~**DO NOT MERGE** until ros2 nightly archives are based on ubuntu focal~~

This does not update the source image. That one cannot be updated until the official foxy release or until a foxy-release tag is created on ros2/ros2 repository.

Question: should the `ros2:nightly` base image be made configurable via build arg like `ros2:devel`?
The only advantage is to be able to support various archive types. But at that point it's likely that some other change in the dockerfile would be necessary so making it a build arg doesn't bring that much

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -1,6 +1,6 @@
 # This is an auto generated Dockerfile for ros:nightly
 # generated from docker_images_ros2/nightly/create_ros_image.Dockerfile.em
-FROM ubuntu:bionic
+FROM ubuntu:focal
 
 # setup timezone
 RUN echo 'Etc/UTC' > /etc/timezone && \
--- a/ros2/nightly/platform.yaml
+++ b/ros2/nightly/platform.yaml
@@ -4,7 +4,6 @@
 platform:
     os_name: ubuntu
     os_code_name: focal
-    rosdistro_name: noetic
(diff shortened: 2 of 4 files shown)


mikaelarguedas:
Yeah I did the same, and all deps are there which is good; but yeah nothing will run becasuse all the deps versions are so far off.

> Where do we find out about this? 

Only info available was "very near future" https://github.com/osrf/docker_images/issues/379#issuecomment-595462460. I couldnt find any PR of issue to watch so I asked [here](https://github.com/osrf/docker_images/issues/379#issuecomment-595464815)

mikaelarguedas:
> I naively tested this locally just now; looks like it builds even though the archive was built using bionic,

Yeah CI actually test builds those on PRs now :tada: 

>  Is there a line in a buildfarm config we could link to here?

It would be this line https://github.com/ros2/ci/blob/d8af044b172a6d56da6ccf219401f6da72fb16a4/linux_docker_resources/Dockerfile#L1

> Where do we find out about this?

From https://github.com/osrf/docker_images/issues/379#issuecomment-595920085
@dirk-thomas will ping us when this can be rolled out

mikaelarguedas:
nightly archives use focal https://github.com/ros2/ci/pull/414, merging this to fix the nightly docker images

ruffsl:
Yeah!

PR #386, opened 2020-03-19
commit docker library only if there are changes

mikaelarguedas:
[CI failed last night](https://github.com/osrf/docker_images/runs/518016036) because some dockerfiles needed to be updated (ros1-bridge) but as the bridge image is not part of the official docker library there was nothing to commit library-wise.

This handles the case where the diff on the lib is null and make the bot open the PR with only the Dockerfile changes if library not impacted.

Example of test job printing: "Would create a PR titled ..." instead of actually opening the PR https://github.com/osrf/docker_images/runs/519280424

### DIFF
--- a/.ci/ci_script.py
+++ b/.ci/ci_script.py
@@ -195,11 +195,14 @@ def main(argv=sys.argv[1:]):
                 '--manifest', manifest,
                 '--output', output))
 
-            # Commit changes to Docker Library
-            repo.git.add(all=True)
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Merging to fix CI and trigger PR bot.

Happy to address comments after the fact if any

mikaelarguedas:
Example of resulting PRs:
https://github.com/osrf/docker_images/pull/389
https://github.com/osrf/docker_images/pull/388

Issue #257, opened 2019-04-08
What license are these Dockerfiles under


cjds:
I was just curious what the License was for this Dockerfile? Other packages in this Group seem to be licensed as BSD (https://github.com/osrf/rosbridge_suite/blob/develop/LICENSE).

moriarty:
@tfoote & @wjwwood & @jacobperron I have the same question as @cjds 

It's easy to extend these dockerfiles with
```dockerfile
FROM ros:melodic
```
```dockerfile
FROM gazebo:gzserver9
```

For basic cases. But if you want to do something more unique like mixing gazebo version with some other ros version, and enabling specific hardware acceleration while minimizing the size of the final docker image you end up needing to write from scratch.

```dockerfile
FROM ubuntu:bionic
```

From this line:
_You can however use this repo to build and customize your own images to fit your software/hardware configuration._

On the [How to use this image](https://registry.hub.docker.com/_/gazebo/) document on docker hub, I assume these are released under a permissive license


> Devices
> As of Gazebo version 5.0, physics simulation under a headless instances of gzserver works fine. However some application may require image rendering camera views and ray traces for other sensor modalities. For Gazebo, this requires a running X server for rendering and capturing scenes. In addition, graphical hardware acceleration is also needed for reasonable realtime framerates. To this extent, mounting additional graphic devices into the container and linking to a running X server is required. In the interest of maintaining a general purpose and minimalistic image which is not tightly coupled to host system software and hardware, we do not include tags here with these additional requirements and instructions. You can however use this repo to build and customize your own images to fit your software/hardware configuration. The OSRF's Docker Hub organization profile contains a Gazebo repo at osrf/gazebo which is based on this repo but includes additional tags for these advanced use cases.

ruffsl:
See suggested fix here: https://github.com/osrf/docker_images/pull/400

PR #400, opened 2020-04-07
Add LICENSE (Apache 2.0)

ruffsl:
It seems that other official docker hub images, like the Ubuntu and Debian image that we depend upon, are released under Apache License v2.0:

* https://github.com/tianon/docker-brew-ubuntu-core/blob/master/LICENSE
* https://github.com/debuerreotype/docker-debian-artifacts/blob/master/LICENSE

Given that precedent, we could follow suit and release our repo under the same license.

Fixes https://github.com/osrf/docker_images/issues/257

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
I'd like a thumbsup from @tfoote before merging.

Apache 2.0 sounds fine to me.
The built images also come with that extra verbiage: https://github.com/docker-library/docs/blob/master/.template-helpers/license-common.md

moriarty:
:+1:

ruffsl:
ping @tfoote

PR #405, opened 2020-05-25
Create Noetic dockerfiles

mikaelarguedas:
These files won't compile just yet, because apparently the ROS bootstrap tools are not released in their python2 version for focal.

We'll need to modify the [bootstrap tools template](https://github.com/osrf/docker_templates/blob/d2b95496aede816e6f33b412b5c20dcae7bb7ac5/docker_templates/templates/snippet/install_ros_bootstrap_tools.Dockerfile.em#L1) to not only use the `ros_version` but to also get the `ROS_PYTHON_VERSION` to install the tools accordingly. It would be great to get it from e.g. rosdistro rather than hard-coding it.

@sloretz FYI

### DIFF
--- a/ros/noetic/debian/buster/ros-base/Dockerfile
+++ b/ros/noetic/debian/buster/ros-base/Dockerfile
@@ -1,0 +1,21 @@
+# This is an auto generated Dockerfile for ros:ros-base
+# generated from docker_images/create_ros_image.Dockerfile.em
+FROM ros:noetic-ros-core-buster
+
+# install bootstrap tools
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    build-essential \
+    python-rosdep \
+    python-rosinstall \
+    python-vcstools \
+    && rm -rf /var/lib/apt/lists/*
+
+# bootstrap rosdep
+RUN rosdep init && \
+  rosdep update --rosdistro $ROS_DISTRO
+
+# install ros packages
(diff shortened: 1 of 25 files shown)


ruffsl:
Is that something we could put in the platform yaml? `ros_python_version`

ros/noetic/debian/buster/ros-base/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# This is an auto generated Dockerfile for ros:ros-base
+# generated from docker_images/create_ros_image.Dockerfile.em
+FROM ros:noetic-ros-core-buster
+
+# install bootstrap tools
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    build-essential \
+    python-rosdep \
```
ruffsl, line 8: Ah, thought you meant the template wouldn't compile the same Dockerfiles, as if you had patched them manually.

mikaelarguedas:
As discussed offline with @ruffsl , right now we monkey-patch the dockerfile by hand to enable the initial release.
 
The better approach will be to use rosdistro to get the missing information. Follow-up tracking ticket at https://github.com/osrf/docker_images/issues/406

mikaelarguedas:
Images built and smoke-tested locally with:

```
source /opt/ros/noetic/setup.bash 
rosrun rospy_tutorials add_two_ints_server &
rosrun roscpp_tutorials add_two_ints_client 3 5
```

PR #422, opened 2020-06-24
Add Rolling

mikaelarguedas:
The first sync of ROS Rolling happened: https://discourse.ros.org/t/new-packages-for-ros-2-rolling-ridley-2020-06-23/15019

A couple notes:
- I put an EOL of 2022, rationale being that any time in the future could work as it has no "support period", however many assumptions of the current set of images won't hold in 2022 when ROS2-H comes out (ubuntu version, bridging with noetic etc) so we'll need to change things in these images at that time
- this is using the same repo (main) as the other images. It's unclear to me how often sync will happen and if the testing repo has much meaning for an unstable/experimental distro. So I kept main for now
Edit: from [REP-2002](https://www.ros.org/reps/rep-2002.html#support-status) it says the syncs will happen regularly just like other distros

### DIFF
--- a/.github/workflows/ros_ci.yaml
+++ b/.github/workflows/ros_ci.yaml
@@ -6,6 +6,7 @@ on:
     paths:
     - '.ci/**'
     - '.github/workflows/ros_ci.yaml'
+    - 'ros/rolling/**'
     - 'ros/eloquent/**'
     - 'ros/dashing/**'
     # - 'ros/noetic/**'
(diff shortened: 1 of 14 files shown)


ruffsl:
Nice addition on the docs, thanks!

mikaelarguedas:
> LGTM. Perhaps we could add a comment on the README our library docs eventually to note the niche this tag fills. Perhaps for development, somewhere in between the stable LTS releases versus the rather fat nightly tags.

Very good point, https://github.com/osrf/docker_images/pull/422/commits/a12e2ed41c2856fb056530f380788a2ec4e0bbcd

PR #430, opened 2020-07-22
EOL tags built from debian stretch

ruffsl:
Related: https://github.com/docker-library/official-images/pull/8405#issuecomment-662167454


mikaelarguedas:
Looks like debs for stretch are not built by the builfarm anymore: https://github.com/ros/rosdistro/pull/27640
Once we have the final snapshot for that distro we will be able to build a last image using the snapshot and retire all the Melodic/Stretch images

mikaelarguedas:
Replaces by #541

PR #454, opened 2020-09-03
test sha256 injection

ipa-mdl:
tests https://github.com/osrf/docker_templates/pull/90

### DIFF
--- a/gazebo/10/ubuntu/bionic/libgazebo10/Dockerfile
+++ b/gazebo/10/ubuntu/bionic/libgazebo10/Dockerfile
@@ -1,7 +1,11 @@
 # This is an auto generated Dockerfile for gazebo:libgazebo10
 # generated from docker_images/create_gzclient_image.Dockerfile.em
 FROM gazebo:gzserver10-bionic
+# label gazebo packages
+LABEL sha256.libgazebo10-dev=c069dd36e141269464acd9b0fb560ce6928aad77b661b4ef811e9b16432467b4
+
 # install gazebo packages
-RUN apt-get update && apt-get install -q -y --no-install-recommends \
+RUN apt-get update && apt-get install -y --no-install-recommends \
     libgazebo10-dev=10.2.0-1* \
     && rm -rf /var/lib/apt/lists/*
+
--- a/ros/dashing/ubuntu/bionic/ros1-bridge/Dockerfile
+++ b/ros/dashing/ubuntu/bionic/ros1-bridge/Dockerfile
@@ -10,12 +10,22 @@ RUN echo "deb http://packages.ros.org/ros/ubuntu bionic main" > /etc/apt/sources
 
 ENV ROS1_DISTRO melodic
 ENV ROS2_DISTRO dashing
+# label ros packages
+LABEL sha256.ros-melodic-ros-comm=a0fca5a04686610280d507c61b76c956a7acada71255910b7b9b47bf44a375dc \
+      sha256.ros-melodic-roscpp-tutorials=d676c108b642f0243992795ee3fb5c75e2ef0b28ccd1e843a9d8603bda3a6555 \
+      sha256.ros-melodic-rospy-tutorials=46ee4d1e9f93cca81e75ef0fdbc6701e90ffab2514f049e96700c9f1114cdcea
+
 # install ros packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
     ros-melodic-ros-comm=1.14.9-1* \
     ros-melodic-roscpp-tutorials=0.9.3-1* \
     ros-melodic-rospy-tutorials=0.9.3-1* \
     && rm -rf /var/lib/apt/lists/*
+
+# label ros2 packages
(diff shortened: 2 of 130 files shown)


gazebo/10/ubuntu/bionic/libgazebo10/Dockerfile
```diff
@@ -1,7 +1,11 @@
 # This is an auto generated Dockerfile for gazebo:libgazebo10
 # generated from docker_images/create_gzclient_image.Dockerfile.em
 FROM gazebo:gzserver10-bionic
+# label gazebo packages
```
ruffsl, line 4: Nit pick: could you tweak the templates so there is always a space before line comments.

ros/dashing/ubuntu/bionic/ros1-bridge/Dockerfile
```diff
@@ -10,13 +10,23 @@ RUN echo "deb http://packages.ros.org/ros/ubuntu bionic main" > /etc/apt/sources
 
 ENV ROS1_DISTRO melodic
 ENV ROS2_DISTRO dashing
+# label ros packages
```
ruffsl, line 13: Same new line nitpick.

ros/lunar/debian/stretch/desktop-full/Dockerfile
```diff
@@ -2,8 +2,11 @@
 # generated from docker_images/create_ros_image.Dockerfile.em
 FROM osrf/ros:lunar-desktop-stretch
 
+# label ros packages
```
ruffsl, line 5: Not sure if we should change EOL Dockerfile. I not against it, as I like consistency, but we should check for side effects.

ros/lunar/debian/stretch/ros-core/Dockerfile
```diff
@@ -14,24 +14,17 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 4B63CF8FDE
...
-    && rm -rf /var/lib/apt/lists/*
-
 # setup environment
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
-# bootstrap rosdep
-RUN rosdep init \
-    && rosdep update --include-eol-distros
+ENV ROS_DISTRO lunar
```
ruffsl, line 21: Not sure if we should change EOL Dockerfile. I not against it, as I like consistency, but we should check for side effects.

ros/lunar/debian/stretch/ros-core/Dockerfile
```diff
@@ -14,24 +14,17 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 4B63CF8FDE
...
+
+# label ros packages
+LABEL sha256.ros-lunar-ros-core=cdf4fa8a555d13e8d69b7bf2fd69d8169d36949d40ee7700a642b9a125f77204
```
ruffsl, line 24: Something @mikaelarguedas mentioned to me. Is this the same sha the is in the snapshot repo?

mikaelarguedas:
> Not sure if we should change EOL Dockerfile. I not against it, as I like consistency, but we should check for side effects.

:+1: I like the ideas that the images not rebuilt by dockerhub anymore dont see changes. The fact that only non-EOL (active in the manifest file) Dockerfile get updated sits well with me.

An easy way to just regenerate those is to use the global gneerate_dockerfolders.py script:
```
root@5f28475d0a7e:/tmp/docker_images/ros# ./create_dockerfolders.py dir -d . -a
root@5f28475d0a7e:/tmp/docker_images/gazebo# ./create_dockerfolders.py dir -d . -a
```

ros/lunar/debian/stretch/ros-core/Dockerfile
mikaelarguedas: If we decide to not update EOL images (suggested in https://github.com/osrf/docker_images/pull/454#issuecomment-689149674) we should end up with all the Dockerfiles using snapshots.ros.org remaining unchanged in this PR.

If we decide to update them anyways, we should check that the sha is matching the one from snapshots.ros.org.

mikaelarguedas:
Unfortunately some Dockerfiles have diverged from the templates for the time being.
Some of the changes we'll need to keep as is and not override:
- EOL images (the ones not generated by the commands in https://github.com/osrf/docker_images/pull/454#issuecomment-689149674)
- installation of the ros bootstrap tools (in noetic) should be kept as `python3-<package_name>`

mikaelarguedas (commented):
This is really great thanks @ipa-mdl for working on this. This looks good and aside from a couple nitpicks it's almost ready for prime time 

I made a couple comments about the files that IMO should not be modified to stay true to their last image built on dockerhub as well as some "not matching the template" installs in noetic that we should keep for the resulting image to stay usable

ruffsl:
Closing for https://github.com/osrf/docker_images/pull/457

PR #458, opened 2020-09-24
Remove workaround for pytest

ruffsl:
Now that cited issues are closed. See:
* https://github.com/ros2/ros2_documentation/issues/271
* https://github.com/ros2/ros2/issues/722

### DIFF
--- a/ros2/source/devel/Dockerfile
+++ b/ros2/source/devel/Dockerfile
@@ -50,10 +50,6 @@ RUN pip3 install -U \
     flake8-quotes \
     pytest-repeat \
     pytest-rerunfailures
-# This is a workaround for pytest not found causing builds to fail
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
These lines act as integration tests as well. They ensure that pytest is installed and found properly. I think there is value in keeping them to catch future breakages early

ruffsl:
> These lines act as integration tests as well. They ensure that pytest is installed and found properly.

Not sure how the current directive halts the image build if the incorrect pytest version is found, vs. if it's not found at all.

> I think there is value in keeping them to catch future breakages early

Ok. Lets let ti be then.

mikaelarguedas:
We had an issue that some versions were incompatible preventing python from finding pytest at all.
This ensures that pytest can be found / invoked meaning there is a "working" version. We do not enforce version number as long as pytest can be used

PR #478, opened 2020-10-13
Update sha256 labels

ruffsl:
See https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md#rules

Related:
* https://github.com/docker-library/official-images/pull/8786#issuecomment-700238992
* https://github.com/osrf/docker_templates/pull/92


ruffsl:
Not sure when patch in e968311 was removed, nor if it's still needed, but I just added it back, as the image template still hadn't removed it.

ros/eloquent/ubuntu/bionic/images.yaml.em
```diff
@@ -43,10 +43,3 @@ images:
             - ros-comm
             - roscpp-tutorials
             - rospy-tutorials
-        downstream_packages:
```
mikaelarguedas, line 46: Is this not necessary anymore ?

ros/eloquent/ubuntu/bionic/ros1-bridge/Dockerfile
```diff
@@ -35,11 +35,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     ros-eloquent-demo-nodes-py=0.8.4-1* \
     && rm -rf /var/lib/apt/lists/*
 
-# install downstream packages
```
mikaelarguedas, line 38: Is this not necessary anymore ?

mikaelarguedas (commented):
> Not sure when patch in e968311 was removed, nor if it's still needed, but I just added it back, as the image template still hadn't removed it.

Looks like you added it back only for dashing but that you removed it for eloquent. Is it on purpose ?

ros/eloquent/ubuntu/bionic/images.yaml.em
ruffsl: I looked at the ros_ci.yaml to remember to check which distro's where patched, but only saw noetic being treated as special. Patch re-applied here: https://github.com/osrf/docker_images/pull/479

ros/eloquent/ubuntu/bionic/ros1-bridge/Dockerfile
ruffsl: https://github.com/osrf/docker_images/pull/478#discussion_r506654208

PR #479, opened 2020-10-16
Repatch eloquent

ruffsl:
Related: https://github.com/osrf/docker_images/pull/478#discussion_r506086790


mikaelarguedas (approved):
thanks!

I confirmed the patch is still necessary

PR #526, opened 2021-04-27
Gazebo 7 and 10 EOL since January 2021

mikaelarguedas:
Announcement https://community.gazebosim.org/t/gazebo-7-10-officially-end-of-life/799


ruffsl:
Do gazebo images need any final adjustments before archiving them, like swapping to snapshot repos or such?

mikaelarguedas:
> Do gazebo images need any final adjustments before archiving them, like swapping to snapshot repos or such?

Not that I know of. So far when retiring gazebo images we just removed them from the build list / library. But Gazebo 7 is the first LTS so maybe something different should be done now ? Like using a more permanent snapshot repo ?
cc @chapulina and @j-rivero

chapulina:
We won't be removing EOL'ed debians from https://packages.osrfoundation.org, so technically the Docker images can continue functioning indefinitely. But we're not going to be maintaining them in any way, if something breaks, it's broken.

So I think it's ok to archive these images :+1:

ruffsl (approved):

mikaelarguedas:
upstream PR: https://github.com/docker-library/official-images/pull/10062

PR #529, opened 2021-05-05
EOL Eloquent

mikaelarguedas:
Eloquent has been EOl since 2020-11.

There doesnt seem to be a final snapshot for it though so we cannot do a final image like we do for other ROS images at the moment.

@tfoote @nuclearsandwich Are ROS2 distros supposed to get a final snapshot when they EOL ?

### DIFF
--- a/.github/workflows/ros_ci.yaml
+++ b/.github/workflows/ros_ci.yaml
@@ -8,7 +8,6 @@ on:
     - '.github/workflows/ros_ci.yaml'
     - 'ros/rolling/**'
     - 'ros/foxy/**'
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
> @tfoote @nuclearsandwich Are ROS2 distros supposed to get a final snapshot when they EOL ?

Since we were snapshotting Eloquent I should do a final one. The snapshot repository host has disk space allocation issues that need to be addressed before I can create new snapshots (which is why there's currently a backlog).

mikaelarguedas:
replaced by #541

Issue #533, opened 2021-05-25
Adding ROS2 testing image tags


ruffsl:
I'm wondering if it would be worth having a `testing` tag targeting the testing repo. Sort of like the `devel` tag, but using the testing repo instead of latest release, and perhaps setting the usual ENVs to rolling. 

https://github.com/osrf/docker_images/blob/75615767101b14e4372ff4c0d129a493733f51df/ros2/source/devel/Dockerfile#L24-L25

``` diff
# setup sources.list
- RUN echo "deb http://packages.ros.org/ros2/ubuntu `lsb_release -sc` main" &gt; /etc/apt/sources.list.d/ros2-latest.list<br class="Apple-interchange-newline">
+ RUN echo "deb http://packages.ros.org/ros2-testing/ubuntu `lsb_release -sc` main" &gt; /etc/apt/sources.list.d/ros2-testing.list<br class="Apple-interchange-newline">
```

Hmm, well, I guess more like `nightly` but without installing rolling from the fat archive or nurfing rosdistro.

We'd probably want to revisit how to keep these images up-to-date.
I think we could use a similar pattern that I've used for Nav2, but add pip update and such:

https://github.com/ros-planning/navigation2/blob/22407b7522b4b07cc65a0c3ab865aa4dc7c34282/.github/workflows/trigger_dockerhub.yaml#L22-L32 

CC @mikaelarguedas @nuclearsandwich

PR #534, opened 2021-05-27
Add ROS2 testing image

ruffsl:
Fixes: https://github.com/osrf/docker_images/issues/533

Useful for CI, where not all of `osrf/ros2:nightly` is needed, but `ros:rolling` is too behind upstream.

CC @SteveMacenski

### DIFF
--- a/.github/workflows/trigger_testing.yaml
+++ b/.github/workflows/trigger_testing.yaml
@@ -1,0 +1,47 @@
+---
+name: CRON Trigger testing images
+
+on:
+  schedule:
+  # 7am UTC, 12am PDT
+  - cron:  '0 7 * * *'
+  push:
+    branches:
+    - main
+    paths:
+    - 'ros2/testing/**'
+
+jobs:
+  testing_build:
+    name: Trigger testing image
+    runs-on: ubuntu-latest
+    container:
+      image: osrf/ros2:testing
+    steps:
+      - name: "Check apt updates"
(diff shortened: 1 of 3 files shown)


mikaelarguedas (commented):
could we generate that from template instead of "hard-coding" all this in the Dockerfile. It looks like most lines are equivalent to the expansion of our already existing templates.

For supporting the testing apt repo we could just add an argument to this template: https://github.com/osrf/docker_templates/blob/8dcaa2592d539a3d715a3886274404eec1bccd66/docker_templates/templates/snippet/setup_ros_sources.Dockerfile.em#L21

ruffsl:
> could we generate that from template instead of "hard-coding" all this in the Dockerfile. It looks like most lines are equivalent to the expansion of our already existing templates.
 
I think this Dockerfile cut acrosses two templates, so we'll have to duplicate the template then.

mikaelarguedas:
> I think this Dockerfile cut acrosses two templates, so we'll have to duplicate the template then.

IMO it is ok to haqve different template as long as they all use the same common snippets. The GPG key incident proved that it was valuable to have a single snippet for setting up the ROS keys rather than having to go modify N Dockerfiles by hand. 

A template based-approach available on https://github.com/osrf/docker_images/pull/542 and https://github.com/osrf/docker_templates/pull/97

If we go that route we'd need to update the CI script accordingly (to expand the template on top of looking for upgrades)

PR #541, opened 2021-06-02
EOL melodic stretch and eloquent

mikaelarguedas:
Now that final images have been built with the final snapshots: https://github.com/docker-library/official-images/pull/10270
we can retire the Eloquent and Melodic on Debian Stretch images

Replaces https://github.com/osrf/docker_images/pull/529 and https://github.com/osrf/docker_images/pull/430


mikaelarguedas:
Upstream PR: https://github.com/docker-library/official-images/pull/10293

PR #542, opened 2021-06-06
Testing alternative

mikaelarguedas:
template based approach to #534 
Requires https://github.com/osrf/docker_templates/pull/97


ruffsl:
> If we go that route we'd need to update the CI script accordingly (to expand the template on top of looking for upgrades)

Did you want also add the testing dockerfile to the CI ROS 2 images workflow?
https://github.com/osrf/docker_images/blob/master/.github/workflows/ros2_ci.yaml

ruffsl:
> Did you also want to add this testing dockerfile to the CI ROS 2 images workflow?

Nevermind, looks like we're not yet calling `.ci/ci_script.py` from the CI ROS 2 images workflow anyhow.

ruffsl (approved):

PR #545, opened 2021-06-11
Dashing final snapshot images

mikaelarguedas:
Dashing [is EOL](https://github.com/osrf/docker_templates/pull/99), this PR switches the images to use the final snapshots before retiring them


ruffsl:
Upstream: https://github.com/docker-library/official-images/pull/10343

PR #568, opened 2021-10-06
Bump ros-comm in foxy,galactic and rolling ros1-bridge images

mikaelarguedas:


mikaelarguedas:
upstream PR: https://github.com/docker-library/official-images/pull/11039

PR #619, opened 2022-05-11
Add rolling desktop-full

chapulina:
Desktop-full is finally on ROS 2, see

* https://github.com/ros-infrastructure/rep/pull/334
* https://github.com/ros2/variants/pull/34

`variants` hasn't been bloomed with the new package yet, but I wanted to make sure I opened this PR before I forgot. The exact hardcoded version will likely change.

### DIFF
--- a/ros/rolling/ubuntu/jammy/desktop-full/Dockerfile
+++ b/ros/rolling/ubuntu/jammy/desktop-full/Dockerfile
@@ -1,0 +1,9 @@
+# This is an auto generated Dockerfile for ros:desktop-full
+# generated from docker_images_ros2/create_ros_image.Dockerfile.em
+FROM osrf/ros:rolling-desktop-jammy
+
+# install ros2 packages
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    ros-rolling-desktop-full=0.9.3-2* \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Great thanks!

Why not creating docker images for `perception` and `simulation` like it was the case in ROS 1 before ?
As this change has been made on REP-2001 that cover all ROS2 distros, this means that these new stacks are available in all ROS 2 distros ? (from REP-2001 it seems that desktop_full is available only un Humble but perception and simulation are available for all distris, is that right ?)

---
Most likely this will be superseeded by a PR adding the new images in a generic way to rolling and all future ros distros.

tfoote:
It would be good to add the other variants as well as base images. We can probably map docker image names to the REP variants.

chapulina:
> The exact hardcoded version will likely change.

Updated to 0.10.0-1, see 

* https://github.com/ros/rosdistro/pull/33173

> Why not creating docker images for perception and simulation like it was the case in ROS 1 before ?

Good point, added in [f26b990](https://github.com/osrf/docker_images/pull/619/commits/f26b990a491c91c212ed9db9ae1cc66240b38d69)

>  from REP-2001 it seems that desktop_full is available only un Humble but perception and simulation are available for all distris, is that right ?

Ouch, I can see the confusion. See https://github.com/ros-infrastructure/rep/pull/353. The answer is that they're all only available from Humble.

> Most likely this will be superseeded by a PR adding the new images in a generic way to rolling and all future ros distros.

Feel free to close this if https://github.com/osrf/docker_images/pull/620 gets merged before, thanks!

mikaelarguedas:
Superseeded by https://github.com/osrf/docker_images/pull/620

Issue #640, opened 2022-10-03
ROS 2 nightly seems to embed wrong version of flake8


Guillaumebeuzeboc:
The ROS 2 nightly docker image seems to be embedding the wrong version of `flake8`.

`flake8` is [installed with PIP3](https://github.com/osrf/docker_images/blob/master/ros2/nightly/nightly/Dockerfile#L50) in the [version 5.0.4](https://pypi.org/project/flake8/).

The flake8 package [available from APT is the version 4.0.1](https://packages.ubuntu.com/jammy/python3-flake8). This package is a dependency of the [python3-ament-cmake](https://packages.ubuntu.com/jammy/python3-ament-flake8).

The package from rolling `ament-lint` doesn't support flake8 5.0.4 since [it's using the option `prog` from the `OptionManager`](https://github.com/ament/ament_lint/blob/rolling/ament_flake8/ament_flake8/main.py#L177:L200) that exist in the [4.0.1](https://github.com/PyCQA/flake8/blob/4.0.1/src/flake8/options/manager.py#L350) but no longer in the [5.0.4](https://github.com/PyCQA/flake8/blob/5.0.4/src/flake8/options/manager.py#L318).

This cause `ament-lint` to crash because it's setting the option `prog` that doesn't exist.

Why is flake8 installed from PIP?
Should the PIP install of `flake8` be removed from the docker image?
(I noticed the same problem with the PIP package `pycodestyle`)

---
To reproduce:
`docker run -ti osrf/ros2:nightly bash`
`mkdir -p ws/src`
`cd ws/src`
`git clone https://github.com/ros2/demos.git`
`cd ../`
`rosdep update`
`rosdep install --from-paths src --ignore-src --rosdistro rolling -y`
`colcon build --event-handlers console_cohesion+ --packages-select topic_monitor`
`colcon test --packages-select topic_monitor --return-code-on-test-failure --event-handlers console_cohesion+`
error: 
```
Starting >>> topic_monitor
--- output: topic_monitor
============================= test session starts ==============================
platform linux -- Python 3.10.6, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
cachedir: /r2_ws/build/topic_monitor/.pytest_cache
rootdir: /r2_ws/src/demos, configfile: pytest.ini
[...]

PR #641, opened 2022-10-04
install flake8 and pytest plugins from apt instead of pip

mikaelarguedas:
Closes https://github.com/osrf/docker_images/issues/640

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -18,7 +18,18 @@ RUN apt-get update && apt-get install -q -y --no-install-recommends \
     gnupg2 \
     libssl-dev \
     lsb-release \
+    python3-flake8 \
+    python3-flake8-blind-except \
+    python3-flake8-builtins \
+    python3-flake8-class-newline \
+    python3-flake8-comprehensions \
+    python3-flake8-deprecated \
+    python3-flake8-docstrings \
+    python3-flake8-import-order \
+    python3-flake8-quotes \
     python3-pip \
+    python3-pytest-repeat \
+    python3-pytest-rerunfailures \
     wget \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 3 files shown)


ros2/nightly/nightly/Dockerfile
```diff
@@ -46,18 +57,7 @@ ENV ROSDISTRO_INDEX_URL https://raw.githubusercontent.com/osrf/docker_images/mas
...
-    flake8-builtins \
-    flake8-class-newline \
-    flake8-comprehensions \
-    flake8-deprecated \
-    flake8-docstrings \
-    flake8-import-order \
-    flake8-quotes \
-    pytest-repeat \
-    pytest-rerunfailures
+    argcomplete
```
ruffsl, line 60: Is there still a reason that argcomplete is soloed out?

Looks like its also available upstream:  
https://packages.ubuntu.com/jammy/python/python3-argcomplete

I couldn't find what the build farm does for this, given argcomplete probably isn't used by the buildfarm.

mikaelarguedas: For a long time the version of argcomplete was too low for several ros2cli and/or colcon features.
Maybe things have evolved since. Looks like in official documentation it switched to apt for that as well: https://github.com/ros2/ros2_documentation/pull/1275/files.

On the other hand if the list of pip packages to install is empty the template will stop installing pip or setuptools altogether that may have side effects so I'd rather not do this here and solely fix the flake8 issue in this PR

ros2/nightly/nightly/Dockerfile
```diff
@@ -18,7 +18,18 @@ RUN apt-get update && apt-get install -q -y --no-install-recommends \
...
     lsb-release \
+    python3-flake8 \
+    python3-flake8-blind-except \
+    python3-flake8-builtins \
+    python3-flake8-class-newline \
+    python3-flake8-comprehensions \
+    python3-flake8-deprecated \
+    python3-flake8-docstrings \
+    python3-flake8-import-order \
+    python3-flake8-quotes \
```
nachovizzo, line 29: @mikaelarguedas , @ruffsl sorry to jump out of nowhere here. But this linter `flake8-quotes` is literally driving me crazy :roll_eyes: 

1. Is there any good reason why we are using it!?
2. Where can I find the original motivation for adding this?
3. Where can I propose to completely get rid of such linter :nerd_face: ?

Thanks in advance and sorry for the annoyance!

mikaelarguedas: Hi @nachovizzo,

This is not something that is decided on the docker images maintaining side.
This is part of the ROS 2 styleguide: https://docs.ros.org/en/humble/The-ROS2-Project/Contributing/Code-Style-Language-Versions.html?highlight=style%20guide#id4 specifically this line:
"We pick single quotes over double quotes as long as no escaping is necessary."

I think the topic came up a couple times, I remember a thread on discourse a while back to compatible with other linters e.g. black https://discourse.ros.org/t/python-black-formatting/13277

Not sur if this is a satisfactory answer but that's a start :)

nachovizzo: @mikaelarguedas 🙌 Thanks! This is what I was missing , this entry point. Thanks a lot

Issue #697, opened 2023-11-20
snapshots.ros.org GPG key expired


patrickwasp:
```docker
FROM ros:foxy

RUN apt-get update
```

building results in:

```
2.558 W: GPG error: http://snapshots.ros.org/foxy/final/ubuntu focal InRelease: The following signatures were invalid: EXPKEYSIG AD19BAB3CBF125EA ROS Spanshot builder <rosbuild@ros.org>
2.558 E: The repository 'http://snapshots.ros.org/foxy/final/ubuntu focal InRelease' is not signed.
------
failed to solve: process "/bin/sh -c apt-get update" did not complete successfully: exit code: 100
```

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy
Docker version 24.0.7, build afdd53b
Docker Compose version v2.21.0

mikaelarguedas:
Thanks for reporting, I have the same issue. This means that all docker images for all EOL ROS Distro are impacted.

@nuclearsandwich @tfoote are you aware of this signature issue ?

gavanderhoorn:
> This means that all docker images for all EOL ROS Distro are impacted.

indeed.

Came to report the same thing, but for ROS 1 docker images.

From one of my CI runs fi (for Melodic):

```
$ gpg --no-options --trust-model always --no-default-keyring --keyring /usr/share/keyrings/ros-archive-keyring.gpg --fingerprint
  /usr/share/keyrings/ros-archive-keyring.gpg
  -------------------------------------------
  pub   rsa3072 2018-11-19 [SC] [expired: 2023-11-19]
        4B63 CF8F DE49 746E 98FA  01DD AD19 BAB3 CBF1 25EA
  uid           [ expired] ROS Snapshot builder <rosbuild@ros.org>
```

jobafr:
This also seems to impact noetic snapshots.

```
http://snapshots.ros.org/noetic/2022-07-27/ubuntu focal InRelease: The following signatures were invalid: EXPKEYSIG AD19BAB3CBF125EA ROS Snapshot builder <rosbuild@ros.org>
```

The normal noetic repo for ubuntu does not seem to be affected.

BrianOpteran:
:+1: This is also afffecting Galactic.

j-rivero:
> @nuclearsandwich @tfoote are you aware of this key expiration ?

[...]

PR #698, opened 2023-11-21
Hack to force gpg key update

mikaelarguedas:
Integrate generic changes from https://github.com/osrf/docker_templates/pull/108 + modify by hand some old images to force a rebuild without changing structure or functionality

Fix for https://github.com/osrf/docker_images/issues/697

### DIFF
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -2,7 +2,7 @@
 ####### requirements.txt #######
 #
 ###### Requirements without Version Specifiers ######
-empy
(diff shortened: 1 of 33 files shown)


ruffsl (approved):
Unsure if the docker librarians would approve the rebuilding of EOL images, given their base images are also EOL. But it's worth asking if they can. I also hope they don't bawk from the diff size, but there's no helping that given the the signing key fingerprint remains unchanged.

Issue #621, opened 2022-05-19
Rolling: apt-key deprecation warning


jaagut:
Hi!

When running `sudo apt update` inside a fresh `ros:rolling` container, apt throws the following warning:

```
W: http://packages.ros.org/ros2/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```

I hope this issue currently does not affect anything, I just wanted to let you know.

Cheers,
Jan

mikaelarguedas:
Thanks @jaagut for reporting it !

This currently has no effect, this is mostly to notify maintainers that for future versions of ubuntu another way of adding apt repository keys should be used.

---

cc @nuclearsandwich @tfoote @ruffsl 
- I see the official ROS2 instructions get the key from `https://raw.githubusercontent.com/ros/rosdistro/master/`. Would you recommend using that appproach in the docker images as well ? One question I have is about availability, there were sometimes issues with keyservers in the past and I wonder if getting files from githubusercontent is robust in your opinion.
- Do you know where we could fetch the snapshots.ros.org key file from ?

nuclearsandwich:
> * Would you recommend using that appproach in the docker images as well ?

Yeah that is currently the most reliable method for fetching the key we're using for the ROS repositories.
We're working on updating the repository hosts and signing key configurations to support the creation of `ros{,2}{,-testing}-archive-keyring and potentially even ros-snapshots-archive-keyring packages as recommended in https://wiki.debian.org/DebianRepository/UseThirdParty so we can more easily rotate repository signing keys with each distribution as well as move away from the use of deprecated repository configuration methods. I was hoping to get this done for Humble's launch but with everything else going on and a strong desire to be deliberate with keys and get things as right as possible the first time that's been pushed out so we don't rush it.

> Do you know where we could fetch the snapshots.ros.org key file from ?

At this time there is no public https server hosting this key. I'd rather not provide one while the eventual goal is to move toward the archive-keyring packages as it would require changing how we do things twice. If you need stable keys for the Docker images I would suggest embedding the key in the Docker images repository and adding it to the image as a local file. i.e.
```Dockerfile
ADD ros-snapshots.asc /tmp/ros-snapshots.asc
RUN apt-key add /tmp/ros-snapshots.asc && rm /tmp/ros-snapshots.asc
```

mikaelarguedas:
[...]

PR #705, opened 2023-12-19
Move to gpg keyring for non-EOL images

mikaelarguedas:
Will close https://github.com/osrf/docker_images/issues/621 and follow guidelines provided in https://github.com/docker-library/official-images/pull/15809#issuecomment-1846053036

Will result in https://github.com/docker-library/official-images/pull/15936

### DIFF
--- a/gazebo/11/ubuntu/focal/gzserver11/Dockerfile
+++ b/gazebo/11/ubuntu/focal/gzserver11/Dockerfile
@@ -16,12 +16,20 @@ RUN apt-get update && apt-get install -q -y --no-install-recommends \
     lsb-release \
     && rm -rf /var/lib/apt/lists/*
 
+
 # setup keys
-RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD83DB69272AFE98867170598AF249743
+RUN set -eux; \
+	key='D2486D2DD83DB69272AFE98867170598AF249743'; \
+	export GNUPGHOME="$(mktemp -d)"; \
+	gpg --batch --keyserver keyserver.ubuntu.com --recv-keys "$key"; \
+	mkdir -p /etc/apt/keyrings; \
+	gpg --batch --export "$key" > /usr/share/keyrings/pkgs-osrf-archive-keyring.gpg; \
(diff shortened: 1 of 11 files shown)


gazebo/11/ubuntu/focal/gzserver11/Dockerfile
```diff
@@ -13,15 +13,22 @@ RUN echo 'Etc/UTC' > /etc/timezone && \
...
 
+
 # setup keys
-RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys D2486D2DD83DB69272AFE98867170598AF249743
+RUN set -eux; \
+	key='D2486D2DD83DB69272AFE98867170598AF249743'; \
+	export GNUPGHOME="$(mktemp -d)"; \
+	gpg --batch --keyserver keyserver.ubuntu.com --recv-keys "$key"; \
+	mkdir -p /usr/share/keyrings; \
+	gpg --batch --export "$key" > /usr/share/keyrings/pkgs-osrf-archive-keyring.gpg; \
```
mikaelarguedas, line 25: @j-rivero I followed the install location from [the gazebo docs](https://gazebosim.org/docs/harmonic/install_ubuntu)

Do you confirm the keyring should be installed in `/usr/share/keyrings` and not in `/etc/apt/keyrings`?
Same for the ROS ones ? (@nuclearsandwich ?)

j-rivero: > Do you confirm the keyring should be installed in `/usr/share/keyrings` and not in `/etc/apt/keyrings`?
> Same for the ROS ones ? (@nuclearsandwich ?)

Following https://wiki.debian.org/DebianRepository/UseThirdParty
> If future updates to the certificate will be managed by an apt/dpkg package as recommended below, then it SHOULD be downloaded into /usr/share/keyrings using the same filename that will be provided by the package. If it will be managed locally , it SHOULD be downloaded into /etc/apt/keyrings instead.

Our intention is to manage the ROS/Gazebo keys using a package )hopefully sometime in the short future). So I'm inclined to use `/usr/share/keyrings` as suggested by the Debian documentation.

PR #706, opened 2024-01-02
attempt to fix dependabot docker version fix

mikaelarguedas:
Apparently recursively finding Dockerfile is not working: https://github.com/dependabot/dependabot-core/issues/1015

Not sure why it stopped working, last successful Prs were from early 2022: https://github.com/osrf/docker_images/pull/594

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,19 @@
 version: 2
 updates:
   - package-ecosystem: "docker"
-    directory: "/ros2"
+    directory: "/ros2/source/devel"
+    schedule:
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
I'll merge this to see if this makes a change and will revert if no improvement is noticed.

It should open PRs to bump ubuntu jammy base image versions

ruffsl:
Huh, that dependabot regression is annoying. Looks like things are moving finally:  
- https://github.com/dependabot/dependabot-core/issues/2178#issuecomment-1979488388

Issue #728, opened 2024-03-16
Update ubuntu base to noble for Rolling


Yadunund:
Rolling has migrated to Ubuntu Noble: https://discourse.ros.org/t/preparing-ros-2-rolling-for-the-transition-to-ubuntu-24-04/35673/16

mikaelarguedas:
I was mislead by that discourse post as well. But the migration is not done yet and multiple variants like simulation and desktop-full [are not available yet](http://repo.ros2.org/status_page/ros_rolling_default.html?q=variant) preventing the generation and release of the docker images. Same thing for the nightly buildfs that are still not targetting rolling (https://ci.ros2.org/job/ci_packaging_linux/)

Hopefully the migration of the main part of ROS (the official variants) will be completed soon !

Yadunund:
Sounds good. Thanks @mikaelarguedas!

MartinPeris:
I am interested in having this working, can I help?

ruffsl:
@mikaelarguedas , from the repo.ros2.org/status_page/ , it looks like core, base, and perception meta packages are now available. Perhaps we could cut a release for just the official library images that only include those tags anyhow? Not sure when desktop and simulation meta packages will ever arrive.

mikaelarguedas:
> from the repo.ros2.org/status_page/ , it looks like core, base, and perception meta packages are now available

Yes, the situation has been the same since https://github.com/osrf/docker_images/issues/728#issuecomment-2002103668, I originally assumed it would be a matter of days. I haven't found any information on the status / progress so I asked here https://discourse.ros.org/t/preparing-ros-2-rolling-for-the-transition-to-ubuntu-24-04/35673/17

> Perhaps we could cut a release for just the official library images that only include those tags anyhow?

Agreed, I tweaked the feature branch accordingly and opened https://github.com/osrf/docker_images/pull/731

I'll move forward and see to get that upstream and we'll update as needed

luca-della-vedova:
Thanks for all the great work! I believe (since the Jazzy image for desktop-full and simulation was released) that there should be no more blockers now to release the remaining `simulation` and `desktop-full` for Rolling on Noble, if that's the case is anyone working on it yet>

mikaelarguedas:
This is currently waiting for the rolling sync.
https://github.com/osrf/docker_images/pull/742#issuecomment-2087567062

Apparently rolling was synced to main just before the `simulation` and `desktop-full` packages were fixed so now we need to wait for the next one...

PR #731, opened 2024-04-07
Migrate rolling images to ubuntu noble

mikaelarguedas:
related to https://github.com/osrf/docker_images/issues/728

Currently this disables the simulation and desktop_full images as these packages are not available on noble yet
(As s side effect it also disables the rolling CI so this can be merged to be released upstream)


mikaelarguedas:
Upstream PR: https://github.com/docker-library/official-images/pull/16551

PR #738, opened 2024-04-26
migrate all osrf/ros2 images to ubuntu noble

mikaelarguedas:
Related to https://github.com/osrf/docker_images/pull/735#issuecomment-2079466268

This has been tested locally by launching the talker_listener example on source and nightly images with all rmw implementations

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -1,5 +1,5 @@
 # multi-stage for setup
-FROM ubuntu:jammy-20240416 as setuper
+FROM ubuntu:noble-20240423 as setuper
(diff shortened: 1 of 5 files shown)


ruffsl (approved):
What happened to argcomplete? Is it finally listed as a rosped depend for the ros2 cli?

mikaelarguedas:
yeah looks like the apt version now has a recent enough version of argcomplete to stop installing it from pip:
https://github.com/ros2/ci/pull/746
https://github.com/ros2/ros2cli/pull/564

Apparently the fallback paths for sros2 to keep working if argcomplete was not installed has also been removed, so it is now a mandatory dependency:
https://github.com/ros2/sros2/pull/289

PR #741, opened 2024-04-30
Add images for Jazzy testing

mikaelarguedas:
@ruffsl curious to know how you feel about this.

Jazzy will be released in a month and is not fully ready yet
Having images available for pre-release would be likely useful

I was thinking of:
- publishing these images (based on the testing apt repo) to dockerhub
  - not add them to CI
  - not move the latest tag
- A jazzy release time:
  - regenerate the images pointing to the right repo
  - move the latest tag
  - submit to dockerhub for rebuild
 
What do you think?

### DIFF
--- a/ros/ros
+++ b/ros/ros
@@ -73,6 +73,28 @@ Directory: ros/iron/ubuntu/jammy/perception
 
 
 ################################################################################
+# Release: jazzy
+
+########################################
+# Distro: ubuntu:noble
+
+Tags: jazzy-ros-core, jazzy-ros-core-noble
+Architectures: amd64, arm64v8
+GitCommit: 543aabeed37ee905ac3737fd38ffbc9894e997db
+Directory: ros/jazzy/ubuntu/noble/ros-core
+
+Tags: jazzy-ros-base, jazzy-ros-base-noble, jazzy
+Architectures: amd64, arm64v8
+GitCommit: 543aabeed37ee905ac3737fd38ffbc9894e997db
+Directory: ros/jazzy/ubuntu/noble/ros-base
+
+Tags: jazzy-perception, jazzy-perception-noble
+Architectures: amd64, arm64v8
+GitCommit: 543aabeed37ee905ac3737fd38ffbc9894e997db
+Directory: ros/jazzy/ubuntu/noble/perception
+
+
+################################################################################
 # Release: rolling
 
 ########################################
--- a/ros/jazzy/ubuntu/noble/Makefile
+++ b/ros/jazzy/ubuntu/noble/Makefile
@@ -1,0 +1,78 @@
+all: help
+
+help:
+	@echo ""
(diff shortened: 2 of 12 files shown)


ruffsl (approved):
Yeah, sounds like a nice plan! I saw your comment on the Jazzy Jalisco Testing and Tutorial Party thread, and agree it would be handy to have some jazzy docker images for folks to play with, as well as run through their own CI pipelines. Thanks for getting ahead on this!

ros/ros
```diff
@@ -72,6 +72,28 @@ GitCommit: bca53bf4c09d771be3ff735da4157203b53ebc2b
...
+
+Tags: jazzy-ros-base, jazzy-ros-base-noble, jazzy
+Architectures: amd64, arm64v8
+GitCommit: 543aabeed37ee905ac3737fd38ffbc9894e997db
+Directory: ros/jazzy/ubuntu/noble/ros-base
+
+Tags: jazzy-perception, jazzy-perception-noble
+Architectures: amd64, arm64v8
+GitCommit: 543aabeed37ee905ac3737fd38ffbc9894e997db
+Directory: ros/jazzy/ubuntu/noble/perception
```
sloretz, line 94: I can confirm [`desktop` works](https://github.com/sloretz/ros_oci_images?tab=readme-ov-file#jazzy-pre-release-images) from the ros2-testing apt repo, and it looks like `desktop_full` is should work now too.

https://repo.ros2.org/status_page/ros_jazzy_default.html?q=variant

mikaelarguedas: Thanks for the heads up! It was not the case when I initially opened this.
Then I'll add it to ci here so that we get the automated build / push of these images as well

mikaelarguedas: It's not clear to me what is now the right place to report this type of issue.
What would you recommend @sloretz ? ros/rosdistro ?

sloretz: > It's not clear to me what is now the right place to report this type of issue.
> What would you recommend @sloretz ? ros/rosdistro ?

That or https://github.com/ros2/ros2

@clalancette FYI Seems like there might be an install issue with `ros-jazzy-simulation`.

clalancette: Yeah, we have to land https://github.com/ros2/variants/pull/42 , and then re-release.  That should unblock this.

sloretz: @clalancette says

> We need to land https://github.com/ros2/variants/pull/42 and https://github.com/ros-infrastructure/rep/pull/401

Sounds like `desktop` works, but `simulation` and `desktop_full` are not yet installable

tfoote:
Great to have this a little bit ahead of time!

ros/ros
mikaelarguedas: :+1:

PR following the new release of the variants for jazzy: https://github.com/osrf/docker_images/pull/742

PR #742, opened 2024-04-30
jazzy 0.10.0-4 -> 0.11.0-1

mikaelarguedas:
Looks like docker images build fail due to a bump in variants https://doi-janky.infosiftr.net/job/multiarch/job/amd64/job/ros/

https://github.com/osrf/docker_images/pull/741#discussion_r1585267598


mikaelarguedas:
@clalancette do we expect the corresponding change to land on main for rolling shortly or will this be in multiple days / weeks ? (in which case I'll do that in a separate PR)

clalancette:
> @clalancette do we expect the corresponding change to land on main for rolling shortly or will this be in multiple days / weeks ? (in which case I'll do that in a separate PR)

We just did a Rolling sync this morning, so it will probably be at least a few days (maybe longer) until we do another one.

mikaelarguedas:
ok thanks, merging and releasing as is then

PR #747, opened 2024-05-08
[ros2][devel/nightly] add some missing pytest dependencies

mikaelarguedas:
Otherwise some jobs based on ros2:devel fail, e.g. if you try to generate coverage for a python package

I took the set of pytest package from the ROS 2 CI:
https://github.com/ros2/ci/blob/3435d04acf17872b07861e81265bc34d7ef40c72/linux_docker_resources/Dockerfile#L120-L125

### DIFF
--- a/ros2/nightly/nightly/Dockerfile
+++ b/ros2/nightly/nightly/Dockerfile
@@ -28,8 +28,12 @@ RUN apt-get update && apt-get install -q -y --no-install-recommends \
     python3-flake8-import-order \
     python3-flake8-quotes \
     python3-pip \
+    python3-pytest-cov \
+    python3-pytest-mock \
     python3-pytest-repeat \
     python3-pytest-rerunfailures \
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
@nuclearsandwich @clalancette not sure if this is the set of packages users need to test packages with pytest ? or if some are not useful and that's why they are not mentionned in the installation instructions ?

clalancette:
> @nuclearsandwich @clalancette not sure if this is the set of packages users need to test packages with pytest ? or if some are not useful for end users and that's why they are not mentionned in the installation instructions ?

That seems like a reasonable set.

They aren't mentioned in the installation instructions because they are only useful if you are running tests, i.e. a "debug"-style installation.  So to try to keep dependencies smaller, we don't have them.

But you can definitely argue that they should be there for a complete set.  If you'd like to open a PR to the documentation to add them, I'd be OK with that.

Issue #753, opened 2024-06-05
Timestamping tags for dependabot usecases 


ruffsl:
It'd be neat to add/rotate the same datetime pattern to the official ROS image tags as Ubuntu to enable dependabot usecases:

- https://github.com/osrf/docker_images/pull/751

Then folks could automate the version bump and rebuild process of their child docker images whenever dependabot detects a new release. Although, perhaps this can already be achieved by pinning the tag via the image digest sha, so maybe not worth it.

```diff
- FROM ros:jazzy-20240605 
+ FROM ros:jazzy@sha256:1c445ce6bd87d8bc884fcd40e8a9d8f7efd939e2e4d321c2c9453d1c3e3ad541
```

mikaelarguedas:
:+1:

I was considering two things regarding timestamping (that could also ensure cache bursting and image rebuild on new ROS syncs) .

- Storing sync date in an environment variable e.g. `ROS_SYNC_DATE` so that users can know from within the environment what version of the ROS stack they are using (something like the [rosdistro tags](https://github.com/ros/rosdistro/tags) stripped from the distro name: `rolling/2024-05-24` -> `2024-05-24`
- Adding the sync date to the tag like done for ubuntu :+1:

WDYT ?

PR #775, opened 2024-10-30
Add ROS_SYNC_DATE to dockerfile busting cache for each sync

mikaelarguedas:
Update generated dockerfiles based on https://github.com/osrf/docker_templates/pull/110


ruffsl:
Related discussion: https://github.com/osrf/docker_images/issues/753#issuecomment-2151415286

ros/humble/ubuntu/jammy/ros-core/Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN echo "deb [ signed-by=/usr/share/keyrings/ros2-latest-archive-keyring.gpg ]
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
+ENV ROS_SYNC_DATE=2024-09-19
```
gavanderhoorn, line 35: Should the name of the env var include something which clarifies it's the repository sync date?

`ROS_SYNC_DATE` seems a bit ambiguous.

mikaelarguedas:
@nuclearsandwich @tfoote what do you think about this approach ?
Some assumptions are made here:
- the fact that a tag `<DISTRO_NAME>/<SYNC_DATE>` exists on rosdistro implies that the packages for that sync are available on the main apt repo
- that the last tag in the sorted list of tags matching that distro is the latest (basically that the format <DISTRO_NAME>/YYYY-MM-DD will be the only one used on rosdistro

ros/humble/ubuntu/jammy/ros-core/Dockerfile
mikaelarguedas: @gavanderhoorn thabnks for the feedback, I was actually looking for feedback on that.

In my mind everytime we speak about "ROS SYNC" it's the sync of the apt repo but its indeed not super explicit.

What do you have in mind ?
something like `ROS_APT_SYNC_DATE` ?

gavanderhoorn: > In my mind everytime we speak about "ROS SYNC" it's the sync of the apt repo

same for me, but that's probably because we're intimately familiar with the ROS release workflow/pipeline.

> something like `ROS_APT_SYNC_DATE` ?

that's already more specific, but perhaps `ROS_PKG_REPOS_SYNC_DATE` would be more self-explanatory? It would also work for non-apt based systems and "package repositories" seems like it would be somewhat intuitively understandable.

ruffsl (approved):

ros/humble/ubuntu/jammy/ros-core/Dockerfile
mikaelarguedas: This was discussed in ROS Infrastructure PMC today. summary and suggestion available here https://github.com/osrf/docker_templates/pull/110/files#r1847080537

mikaelarguedas: We went with `ROSDISTRO_PKGS_SYNC_DATE`

mikaelarguedas:
upstream PR: https://github.com/docker-library/official-images/pull/17960

PR #776, opened 2024-10-30
ENV and ARG should use key=value format

mikaelarguedas:
Resulting images after applying https://github.com/osrf/docker_templates/pull/111


ruffsl (approved):
LGTM, as per the depiction warnings for build check lint
- https://docs.docker.com/reference/dockerfile/#env

PR #791, opened 2025-02-10
Revert ros sync date

ruffsl:
Reverts:
- https://github.com/osrf/docker_images/pull/775

Also updates rolling images from upstream backlog.

Related:
- https://github.com/osrf/docker_templates/pull/112

References:
- https://github.com/docker-library/official-images/pull/17960#issuecomment-2573702862


mikaelarguedas:
Iron fails because it is now EOL.
Mergin the revert as is and making follow-up PR for iron EOL

PR #801, opened 2025-05-17
Disable iron that is EOL

mikaelarguedas:
Now that the final image is available targetting snapshots: https://github.com/docker-library/official-images/pull/18594


mikaelarguedas:
upstream PR: https://github.com/docker-library/official-images/pull/19048

Issue #807, opened 2025-06-01
docker images use invalid key (expired 2025-06-01): `humble-ros-base`, `kilted-ros-base`, `rolling-ros-base`, ...


Ben-grmbl:
The current image available through dockerhub use a ROS repository key that expired on 2025-06-01

To reproduce:

```
FROM ros:humble-ros-base
RUN apt-get update -y
```

```
3.893 W: GPG error: http://packages.ros.org/ros2/ubuntu jammy InRelease: The following signatures were invalid: EXPKEYSIG F42ED6FBAB17C654 Open Robotics <info@osrfoundation.org>
3.893 E: The repository 'http://packages.ros.org/ros2/ubuntu jammy InRelease' is not signed.
```

Same for kilted, rolling, ...
```
1.337 W: GPG error: http://packages.ros.org/ros2/ubuntu noble InRelease: The following signatures were invalid: EXPKEYSIG F42ED6FBAB17C654 Open Robotics <info@osrfoundation.org>
1.337 W: The repository 'http://packages.ros.org/ros2/ubuntu noble InRelease' is not signed.
```

The issue seems related to this pull request, that was merged yesterday, which at first glance would fix this : #805 

However, the current images still appear to be using the old key. Possibly this is because the docker images still need to be updated manually, as per #784?

I tried to work around this by following this guide:
https://discourse.ros.org/t/ros-signing-key-migration-guide/43937/1
as well as this one from 5 years ago: https://discourse.ros.org/t/ros-gpg-key-expiration-incident/20669

However, I was unable to get it to work inside my docker file. Is there a known workaround?

mikaelarguedas:
Thanks for reporting the issue!

> However, the current images still appear to be using the old key.

Yes this change still needs to be merged into the official docker library and all images to be rebuilt: https://discourse.ros.org/t/ros-signing-key-migration-guide/43937/22


> However, I was unable to get it to work inside my docker file. Is there a known workaround?

The following workaround can be used:

## ROS 2 images

### Remove existing keys / source

```
rm /etc/apt/sources.list.d/ros2-latest.list
rm /usr/share/keyrings/ros2-latest-archive-keyring.gpg
```

### Install the new key

```
apt update && apt install -y curl ca-certificates
export ROS_APT_SOURCE_VERSION=$(curl -s https://api.github.com/repos/ros-infrastructure/ros-apt-source/releases/latest | grep -F "tag_name" | awk -F\" '{print $4}') ;\
[...]

PR #815, opened 2025-06-03
Remove noetic from automated build to allow rebuilding all non EOL ROS 2 images

mikaelarguedas:
Related to https://github.com/osrf/docker_images/issues/807

Now that https://github.com/docker-library/official-images/pull/19162 has merged. The "ros" images are being rebuilt.
For the osrf/ros children images to be rebuilt and pushed we need passing CI and updated docker library.

This PR temporarily rules out noetic (that became EOL in the middle of the gpg key expiration event).

It needs to be reverted as soon as there is a working path forward to build a set of ROS Noetic images based on the snapshots repositories before disabling the build definitively

### DIFF
--- a/.github/workflows/ros_ci.yaml
+++ b/.github/workflows/ros_ci.yaml
@@ -10,7 +10,7 @@ on:
     - 'ros/jazzy/**'
     - 'ros/kilted/**'
     - 'ros/humble/**'
-    - 'ros/noetic/**'
+    # - 'ros/noetic/**'
   push:
     paths:
     - '.ci/**'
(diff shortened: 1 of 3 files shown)


ruffsl (approved):

ruffsl:
> For the osrf/ros children images to be rebuilt and pushed we need passing CI and updated docker library.

Later should we make that CI more piecewise? Or keep it atomic as a whole?

ruffsl:
> The combination of active repo expiring key, distro going EOL and snapshots key not being renewed at the same time seem (hopefully) like a rare scenario

The planets really aligned there for maximum catastrophe. :fire: :fire_engine: 
Hopefully this new key distribution approach allows smoother key rotation...

PR #819, opened 2025-06-10
[ROS] retire eol images

mikaelarguedas:
Now that the images have been rebuilt with the updated GPG key we can retire all Ubuntu Focal based ROS images

Follow-up of https://github.com/osrf/docker_images/pull/817 and https://github.com/docker-library/official-images/pull/19221#issuecomment-2957046040


mikaelarguedas:
the rolling package version bump will be done in a separate PR

PR #829, opened 2025-07-25
Fix ros2/nightly images

mikaelarguedas:
Since 24 January 2025 (https://github.com/osrf/docker_images/actions/runs/12944907207) the ROS 2 nightly image doesnt build because the nightly archive doesnt provide a package.xml for fastdds.

This PR does the following:
- ignore fastdds for rosdep install calls in all nightly images
- update the RTI Connext version for rosdep keys and env var paths
- fix docker linter violations regarding casing and ENV/ARG usage


fujitatomoya (commented):
@mikaelarguedas thanks for the fix 👍

PR #836, opened 2025-08-25
CI platform matrix expansion

a1cd:
This pull request updates the GitHub Actions CI workflows to add support for running jobs on both `ubuntu-latest` and `ubuntu-24.04-arm` platforms. This change ensures that all relevant workflows are tested on both standard and ARM-based Ubuntu environments, improving cross-platform compatibility.

### DIFF
--- a/.github/workflows/trigger_testing.yaml
+++ b/.github/workflows/trigger_testing.yaml
@@ -59,7 +59,12 @@ jobs:
     needs:
       - check_files
       - check_image
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+          - ubuntu-latest
+          - ubuntu-24.04-arm
+    runs-on: ${{matrix.platform}}
     steps:
       - name: Checkout
         uses: actions/checkout@v5


a1cd:
Nope, I wrote that code. I think AI could probably have done it at a higher quality. I just have a use case that requires a non-amd64 (aka x86_64) cpu, and would benefit from having the arm64 architecture images available.

ruffsl:
> but no I was testing it on my own actions in my repo.

Ah, can't blame you for pushed-based development with GitHub Actions. It is indeed tricky to test workflows locally.

> I just have a use case that requires a arm64 image, and your image only supports amd64 (aka x86_64) architectures.

The Official Library images for ROS on DockerHub are indeed Multi-platform builds with `arm64v8` support:

https://github.com/docker-library/official-images/blob/8d52c53042ed840fdc468d321c5248876bfc684e/library/ros#L76-L77

Are you instead wishing to use the `osrf/ros` images for the `desktop` tags? Indeed those are still only `amd64`.

ruffsl:
A fair ask. @mikaelarguedas, this org has access to free arm builders for foss projects now, right?
I'll have to check if this'll just work^tm as expected for multi-platform registry manifests.

PR #879, opened 2026-05-08
Disable rolling until migration to resolute complete

mikaelarguedas:
Rolling images currently fail to build. Rolling is now targetting resolute but all packages are not available on main yet.

This PR disables rolling image generation and CI for now.

Follow-up is tracked here: https://github.com/osrf/docker_images/issues/878

This also regenerates the docker library to update commit hashes

### DIFF
--- a/ros/ros
+++ b/ros/ros
@@ -97,7 +97,7 @@ Directory: ros/lyrical/ubuntu/resolute/perception
 
 Tags: rolling-ros-core, rolling-ros-core-noble
 Architectures: amd64, arm64v8
-GitCommit: fd0c0c943e25f48fe199593597ea21d3f3f5935d
+GitCommit: 8cf2903c0f8813aacd3042c71d4d2d56d5068ad5
 Directory: ros/rolling/ubuntu/noble/ros-core
 
 Tags: rolling-ros-base, rolling-ros-base-noble, rolling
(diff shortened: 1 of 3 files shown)


ros/ros
```diff
@@ -9,7 +9,7 @@ GitRepo: https://github.com/osrf/docker_images.git
 
 Tags: humble-ros-core, humble-ros-core-jammy
 Architectures: amd64, arm64v8
-GitCommit: eb5634cf92ba079897e44fb7541d3b78aa6cf717
+GitCommit: 6baec8a7ddecbfeb424ab8e9f8f81a5bce82ba4d
```
ruffsl, line 12: Would we also hold off on updating the manifest for rolling until it build here? Should we bump this here and now, or would that block adding the Lyrical tag to the library if Rolling took longer to recover. Guess it's a no-op as we could omit it from the docker-library manifest all the same.

ruffsl (approved):

ros/ros
mikaelarguedas: good point, I was nos necessarily planning on submitting upstream PR for this right now.

(I actually opened it with only the Lyrical changes https://github.com/docker-library/official-images/pull/21417)

But you're right, I will revert that part and update commit hash once rolling is back to building on the new platform :+1:

mikaelarguedas: af229c647a4b57d95fd32e504efa42e5754f3771

PR #880, opened 2026-05-09
Fail early on deb download failure and use binary mode for sha check

mikaelarguedas:
Based on @tianon feedback from https://github.com/docker-library/official-images/pull/21417#pullrequestreview-4255672560

Note to self: revert https://github.com/osrf/docker_images/commit/0b9b30eedc03500533502d178697bf6889ef2e3e once approved and matching template branch approved


ruffsl (approved):
Neat! TDIL

